PR #21, opened 2020-02-25
Add additional labels/annotations for OpenShift support

edavidj:

- See [#175](https://github.com/appsody/appsody-operator/issues/175) from appsody for details on why these labels are necessary.
- Added two new labels `app.kubernetes.io/part-of` (defaulting to the app name), and `app.kubernetes.io/component` (defaulting to "backend").
  - This would include all recommend labels, of the optional ones I did not see any others currently missing that we would have information about in application-stacks operator.
  - I was not able to add the recommended annotations from the above link here, as they corresponded to git repository information which, to my understanding, we do not have information on in any scenario with this operator. (If we want them anyways with the default `unknown` I can add that)

- [x] User guide
- [x] `CHANGELOG.md`

### DIFF
--- a/doc/user-guide.md
+++ b/doc/user-guide.md
@@ -158,7 +158,7 @@ for an `RuntimeApplication` CR:
 | `app.kubernetes.io/name`       | `metadata.name`                | A name that represents this component.                                                                                                               |
 | `app.kubernetes.io/managed-by` | `application-runtime-operator` | The tool being used to manage this component.                                                                                                                |
 | `app.kubernetes.io/component`  | `backend`                      | The type of component being created. See OpenShift [documentation](https://github.com/gorkem/app-labels/blob/master/labels-annotation-for-openshift.adoc#labels) for full list. |
-| `app.kubernetes.io/part-of`    | `metadata.name`                | The service that this component belongs to. Configure this if the component is not a standalone application. |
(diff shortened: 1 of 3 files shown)


doc/user-guide.md
```diff
@@ -149,7 +149,21 @@ If applications require specific permissions but still want the operator to crea
...
+| `app.kubernetes.io/instance`   | `metadata.name`                | A unique name or identifier for this component. This cannot be modified.                                                                                                                 |
+| `app.kubernetes.io/name`       | `metadata.name`                | A name that represents this component.                                                                                                               |
+| `app.kubernetes.io/managed-by` | `application-runtime-operator` | The tool being used to manage this component.                                                                                                                |
+| `app.kubernetes.io/component`  | `backend`                      | The type of component being created. See OpenShift [documentation](https://github.com/gorkem/app-labels/blob/master/labels-annotation-for-openshift.adoc#labels) for full list. |
+| `app.kubernetes.io/part-of`    | `metadata.name`                | The service that this component belongs to. Configure this if the component is not a standalone application. |
```
leochr: It might be good to avoid using `service`. Perhpas something like:
_The name of the higher-level application this component is part of_

edavidj: Awesome, I'll go with your suggestion there then 👍

PR #27, opened 2020-03-02
Consistent binding scheme across namespaces

yharish991:

- If the namespace is not provided in the `consumes`, then the secret mount path will not have the namespace in it's path.

* Previously:
If 
```
    consumes:
    - category: openapi
      name: service1
      mountPath: /etc/secrets
```
Pattern - `<mountPath>/<namespace>/<service_name>`
Then secret mount path `/etc/secrets/{namespace}/service1`

* Now, with this PR
If 
```
    consumes:
    - category: openapi
      name: service1
      mountPath: /etc/secrets
```
Pattern - `<mountPath>/<service_name>`
Then secret mount path `/etc/secrets/service1`

- [x] User guide
- [ ] `CHANGELOG.md`

Fixes # https://github.com/application-stacks/operator/issues/23


pkg/apis/appstacks/v1beta1/runtimecomponent_types.go
```diff
@@ -555,7 +568,9 @@ func (cr *RuntimeComponent) Initialize() {
 	for i := range cr.Spec.Service.Consumes {
 		if cr.Spec.Service.Consumes[i].Category == common.ServiceBindingCategoryOpenAPI {
 			if cr.Spec.Service.Consumes[i].Namespace == "" {
-				cr.Spec.Service.Consumes[i].Namespace = cr.Namespace
+				// add "_" suffix to the namespace as a flag
+				// to identify that namespace is not provided in the consumes
+				cr.Spec.Service.Consumes[i].Namespace = cr.Namespace + "_"
```
navidsh: Suffixing `_` works but maybe not ideal since `consumes.namespace` will have an invalid value to users just reading the CR and not knowing the internals. Also ideally we want to keep the type interface as simple as possible (getter funcs).

How about just not initializing `cr.Spec.Service.Consumes[i].Namespace` here if it is not explicitly set?

yharish991: @navidsh That is exactly what i wanted to do initially, but the ReconcileProvides method gets the creates a secret with pattern `namespace-servicename` and if we don't initialize the namespace to anything at the consumer side, ReconcileConsumes method uses namespace to search for the provider secret https://github.com/application-stacks/operator/blob/master/pkg/utils/reconciler.go#L428, if the namespace is empty then it looks for `-servicename` secret which it can never find, needing to explicitly mention the namespace in the consumes even if both the provider and consumer are in the namespace.

Also i wanted the solution to be additive, not break the existing functionality. I agree with your concerns of users reading the CR directly. Do you have any suggestions/ideas on this?

navidsh: @yharish991 What do you think about?
* using the consumer side CR's namespace (`mObj.GetNamespace()`) if `con.GetNamespace()` is empty
* otherwise, use `con.GetNamespace()` as before

https://github.com/application-stacks/operator/blob/75f28f433075b20f68c795d6c9eeb565cfc421cf/pkg/utils/reconciler.go#L427-L429 

There might be other places we need to a similar code change

yharish991:
@navidsh i made the changes according to your suggestions

navidsh:
@yharish991 This looks good. Thanks
Since we haven't added tests yet for the service binding yet, can you run it locally to make sure it's working as expected?

yharish991:
@navidsh yes i tested it locally with both the provider and consumer in same namespace as well as in difference namespace. It works fine.

Issue #3, opened 2020-02-03
Enhance umbrella application support


arthurdm:
We currently have a [mechanism](https://github.com/application-runtimes/operator/blob/master/doc/user-guide.md#kubernetes-application-navigator-kappnav-support) that exploits kAppNav by allowing the auto-creation of `Application` CRs.

This issue proposes an enhancement:
- add a new key to the `RuntimeComponent` CR, called `applicationName`
- if `applicationName` is set to a name that matches an existing `Application` definition, then add the labels to our deployment to `join` it.
- if `applicationName` is not set in any existing `Application` definitions, then add the labels/annotation to our deployment to `create` it (i.e. the same behaviour as `createAppDefinition` today)
- add the appropriate label to the deployment to group it in OCP's Topology view, which should be `app.kubernetes.io/part-of=<appname>` as per [this doc](https://docs.openshift.com/container-platform/4.2/applications/application-life-cycle-management/odc-viewing-application-composition-using-topology-view.html#odc-labels-and-annotations-used-for-topology-view_viewing-application-composition-using-topology-view). 

Some considerations:
- the operator can first check if the `Application` CRD has been installed
- we can preserve the current default behaviour by making `applicationName`'s default value equal `metadata.name`
- keep the key `createAppDefinition` (default remains true) as a toggle to turn off this functionality.

cvignola:
FYI, the kAppNav for labeling application for Application CR auto-creation is here:  https://github.com/kappnav/design/blob/master/auto-app-lifecycle.md

navidsh:
Hi @arthurdm and @cvignola,

To join an existing App definition, if the operator has to add labels to the resources created by the operator, it must "calculate" a set of labels that satisfy all the conditions in the `selector` field of the `Application`.

In some cases, there might be multiple sets of labels that would satisfy the `selector`. In the following example, the operator should add either `tier: business` or `tier: data` labels:

```yaml
kind: Application
...
spec:
  selector:
    matchLabels:
      component: backend
      app: billing
    matchExpressions:
      - {key: tier, operator: In, values: [business, data]}
      - {key: environment, operator: NotIn, values: [dev]}
```

Instead of adding labels to resources created by the operator, how about the operator modifying `Application`'s `selector` field to include one of the labels added to resources by the operator?

arthurdm:
[...]

PR #34, opened 2020-03-04
Adds support for joining existing Applications

navidsh:
Signed-off-by: Navid Shakibapour <navid.shakibapour@gmail.com>

- Adds supports for joining existing Applications CRs by querying their selector labels and adding them to `runtime-component` CR's top-level label
- Always adds `app.kubernetes.io/part-of: <app-name>`

- [ ] User guide
- [ ] `CHANGELOG.md`

Doc related changed to be added in a separate PR.


navidsh:
Fixes https://github.com/application-stacks/runtime-component-operator/issues/3

pkg/controller/runtimecomponent/runtimecomponent_controller.go
```diff
@@ -322,6 +322,20 @@ func (r *ReconcileRuntimeComponent) Reconcile(request reconcile.Request) (reconc
 		return reconcile.Result{}, nil
 	}
 
+	isApplicationSupported, err := r.IsGroupVersionSupported(servingv1alpha1.SchemeGroupVersion.String())
```
arturdzm: I would suggest exploring option to move this to base reconciler, so it can be shared between controllers if possible

pkg/utils/reconciler.go
```diff
@@ -700,3 +700,41 @@ func (r *ReconcilerBase) GetRouteTLSValues(ba common.BaseApplication) (key strin
...
+	var err error
+	if err = r.GetClient().Get(context.Background(), key, app); err == nil {
+		if app.Spec.Selector != nil {
+			for name, value := range app.Spec.Selector.MatchLabels {
+				allSelectorLabels[name] = value
+			}
+		}
+	} else if err != nil && kerrors.IsNotFound(err) {
+		apps := &applicationsv1beta1.ApplicationList{}
+		if err = r.GetClient().List(context.Background(), apps, client.InNamespace("")); err == nil {
```
arturdzm: If intention for ApplicationName to be used as a name of app to join, we should be checking that here as well. It could be that there are 2+ apps that have `kappnav.component.namespaces`  annotation with same NS list. And I think this code will copy  the labels from all of them

navidsh: Good catch! 🥇

navidsh:
@arturdzm Made changes based on the comments

PR #38, opened 2020-03-06
First draft of OLM artifacts

arthurdm:
Added the OLM artifacts by copying the ones from Appsody and changing the references to Runtime Component.  It needs to be updated before releasing, to add the things that went into this repo lately.  The certified CSV will eventually have its registry changed, but I kept the Docker Hub registry for now, until we know the RH registry.

The part I heavily edited was the description, to try to justify the request of maturity level 5 (Auto Pilot).  In the eventual PR to OperatorHub.io we should also mention that this Operator builds on the experience and customer feedback that we've had from the Appsody and Open Liberty Operators.


arthurdm:
I also left the icon as the Appsody one - we'll need to think about a suitable / neutral one.

PR #49, opened 2020-03-09
Adds user-guide for Application support enhancements

navidsh:
Signed-off-by: Navid Shakibapour <navid.shakibapour@gmail.com>

- Adds doc for Application support enhancement

- [x] User guide
- [ ] `CHANGELOG.md`

### DIFF
--- a/doc/user-guide.md
+++ b/doc/user-guide.md
@@ -53,7 +53,7 @@ Each `RuntimeComponent` CR must at least specify the `applicationImage` paramete
 | `serviceAccountName` | The name of the OpenShift service account to be used during deployment. |
 | `applicationImage` | The Docker image name to be deployed. On OpenShift, it can also be set to `<project name>/<image stream name>[:<tag>]` to reference an image from an image stream. If `<project name>` and `<tag>` values are not defined, they default to the namespace of the CR and the value of `latest`, respectively. |
 | `applicationName` | The name of the application this resource is part of. If not specified, it defaults to the name of the CR. |
-| `createAppDefinition`   | A boolean to toggle the automatic configuration of `RuntimeComponent`'s Kubernetes resources to allow creation of an application definition by [kAppNav](https://kappnav.io/). The default value is `true`. See [Application Navigator](#kubernetes-application-navigator-kappnav-support) for more information. |
+| `createAppDefinition`   | A boolean to toggle the automatic configuration of Kubernetes resources for the `RuntimeComponent` CR to allow creation of an application definition by [kAppNav](https://kappnav.io/). The default value is `true`. See [Application Navigator](#kubernetes-application-navigator-kappnav-support) for more information. |
 | `pullPolicy` | The policy used when pulling the image.  One of: `Always`, `Never`, and `IfNotPresent`. |
 | `pullSecret` | If using a registry that requires authentication, the name of the secret containing credentials. |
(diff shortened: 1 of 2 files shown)


navidsh:
Fixes #3

doc/user-guide.md
```diff
@@ -509,9 +510,11 @@ To configure secure HTTPS connections for your Knative deployment, see [Configur
...
+To join an existing application definition in the `RuntimeComponent` CR namespace, ensure the `applicationName` parameter is set to the name of the `Application` CR you want to join to. To join an existing application definition in another namespace, ensure that the `createAppDefinition` parameter is set to `false` and the `applicationName` parameter is set to the name of the existing `Application` CR you want to join to.
+
+The operator tries to find an `Application` CR with the `applicationName` parameter in the `RuntimeComponent` CR namespace. If it fails to find any, then it lists all the `Application` CRs with the value of the `applicationName` parameter as the name and the `RuntimeComponent` CR namespace listed in the value of the `kappnav.component.namespaces` annotation. If it finds any `Application` CR with the specified criteria, it adds labels listed in the `spec.selector.matchLabels` parameter to the `RuntimeComponent` CR. When all these steps fails, if the `createAppDefinition` is not set to `false`, the operator configures the Kubernetes resources it generates to allow automatic creation of an `Application` definition.
```
chirp1: This sentence is too long as it is 44 words: "If it fails to find any, then it lists all the `Application` CRs with the value of the `applicationName` parameter as the name and the `RuntimeComponent` CR namespace listed in the value of the `kappnav.component.namespaces` annotation." We want to break it up so that it is more readable for customers. How about something like: "If it fails to find any, then it lists all the `Application` CRs. The value of the `applicationName` parameter is the name. The `RuntimeComponent` CR namespace listed in the value of the `kappnav.component.namespaces` annotation is the namespace for the Application` CRs". Feel free to slack me so that we come up with the correct wording.

Change " When all these steps fails," to " When all these steps fail,".

doc/user-guide.md
```diff
@@ -569,9 +572,9 @@ spec:
     certificate: {}
 ```
 
-In this scenario the operator will generate `Certificate` resource with common name of `myapp.test.svc` that can be used for service to service communication.
+In this scenario the operator will generate `Certificate` resource with common name of `my-app.test.svc` that can be used for service to service communication.
```
chirp1: Change to "
In this scenario,  the operator generates the `Certificate` resource with a common name of `myapp.test.svc`...".
Change to "In this scenario the operator generates `Certificate` resource with the common name of `my-app.test.svc`...".

doc/user-guide.md
```diff
@@ -569,9 +572,9 @@ spec:
...
 
-Once this certificate request is resolved by cert-manager the resulting secret `myapp-svc-tls` will be 
+Once this certificate request is resolved by cert-manager the resulting secret `my-app-svc-tls` will be
```
chirp1: Change to "After this certificate request is resolved by the certificate manager, the resulting `my-app-svc-tls` secret  is mounted onto each pod inside the `/etc/x509/certs` folder. Mounted files are always up to date with a secret.".

doc/user-guide.md
navidsh: Discussed this through email and fixed in https://github.com/application-stacks/runtime-component-operator/pull/49/commits/b025566b79afd4799209156233426b295f655067.

PR #60, opened 2020-03-23
Add nodePort field to runtime operator

### LINKED ISSUE #53: Ability to set specific NodePort
## Feature Request
Currently the operator sets a random port for the service when using NodePort. As @leochr mentioned NodePort shouldn’t be used in production at all, but this could be used for testing.
Issue [#720](https://github.com/appsody/appsody/issues/720)

### Is your feature request related to a problem?
Could this be something that could be added to the CRD?

yharish991:
* Adds the targetPort field

- [x] User guide
- [ ] `CHANGELOG.md`

Fixes #53


yharish991:
If the service type is `NodePort`, the Kubernetes control plane allocates a port from a range specified by --service-node-port-range flag (default: 30000-32767). So in this PR i added the port range as `1-65535`, this should allow the entire valid port range to accommodate the range set by `--service-node-port-range` flag explicitly by the users. However if the user does not set the `--service-node-port-range` flag and specifies the `nodePort` value that is not in the default range(30000-32767), they should see an error in the operator logs.

pkg/apis/appstacks/v1beta1/runtimecomponent_types.go
```diff
@@ -75,6 +75,10 @@ type RuntimeComponentService struct {
 	// +kubebuilder:validation:Minimum=1
 	TargetPort *int32 `json:"targetPort,omitempty"`
 
+	// +kubebuilder:validation:Maximum=65535
+	// +kubebuilder:validation:Minimum=1
+	NodePort int32 `json:"nodePort,omitempty"`
```
arturdzm, line 80: I think setting this constraints will not allow user switch back to ClusterIP type as `nodePort` cannot be null in this case
Also once set to specific value, there is no way to make it nil again `0` or `nil` values will be not accepted. So there is no way to to switch back to auto-generated port number.
So I think the 0 should be allowed

yharish991:
@arturdzm i agree with what you said, i changed nodePort range to include 0. Actually the problem  you mentioned already exists in the current version of the operator, you can't switch back to `clusterIP` after you have used the `NodePort` service type. This is the error i get when i try to switch
```
Service \"svc1\" is invalid: spec.ports[0].nodePort: Forbidden: may not be used when `type` is 'ClusterIP'
```
With the change i pushed, we should not see this anymore

Issue #69, opened 2020-04-07
Adding support for external service bindings


navidsh:
Tools like `odo` may adopt `ServiceBinding` CR as the resource it creates for service. In that case, users can create the binding themselves and our operator should be able to point to the Service Binding CRs created externally:

```yaml
spec:
  bindings:
    autoDetect: <true or false>
    resourceRef: <name of a manually created ServiceBinding CR>
```

More details on the PR: https://github.com/application-stacks/runtime-component-operator/pull/76

navidsh:
Delivered. Closing.

PR #76, opened 2020-04-14
Added support for external service bindings

navidsh:
- Updates to CRD:
  ```yaml
  spec:
    bindings:
      autoDetect: <true or false>
      resourceRef: <name of a manually created ServiceBinding CR>
  ``` 
- Added a new field to `.status` for keeping track of resolved `ServiceBinding` CRs as `[]string`. For each of the bindings, it adds a `envFrom`. This means all keys from external binding secrets are injected into the container as environment variables.

Notes:
- When there are multiple backing services, a user should combine them into one `ServiceBinding` CR. 
- `resourceRef` has precedence over `audoDetect`, meaning if the value of `resourceRef` is going to be used 

- [ ] User guide
- [ ] `CHANGELOG.md`

Fixes #


navidsh:
Fixes https://github.com/application-stacks/runtime-component-operator/issues/69

PR #80, opened 2020-04-21
Added support for multiple svc ports

### LINKED ISSUE #33: Support multiple service ports
### Is your feature request related to a problem?

A few ppl have asked me how to set two service endpoints with the Operator.  Currently we only allow 1.  

### Describe the solution you'd like

We should allow for an array of service ports / names.

JusteenR:

- Supports the user adding multiple ports to a service
- Allow for an array of service ports/names

- [x] User guide
- [ ] `CHANGELOG.md`

Fixes #33


pkg/utils/utils.go
```diff
@@ -611,7 +651,7 @@ func requiredFieldMessage(fieldPaths ...string) string {
 }
 
 // CustomizeServiceMonitor ...
-func CustomizeServiceMonitor(sm *prometheusv1.ServiceMonitor, ba common.BaseComponent) {
+func CustomizeServiceMonitor(sm *prometheusv1.ServiceMonitor, ba common.BaseComponent, svc *corev1.Service) {
```
arturdzm, line 654: No need to pass svc, all info should be available in ba.GetService()

PR #81, opened 2020-04-22
Support watching service binding secret

navidsh:
- Adding code to watch service binding secrets and find applications that could be relying on the secret
- Updated Role/ClusterRole to include ServiceBinding resource

- [ ] User guide
- [ ] `CHANGELOG.md`

Fixes #


pkg/controller/runtimecomponent/runtimecomponent_controller.go
```diff
@@ -462,10 +486,20 @@ func (r *ReconcileRuntimeComponent) Reconcile(request reconcile.Request) (reconc
 			return r.ManageError(err, common.StatusConditionTypeReconciled, instance)
 		}
 
+		if r.IsOpenShift() {
```
navidsh, line 489: We already fixed this in Appsody Operator but not here

navidsh:
Fixes #69

PR #92, opened 2020-04-27
Update documentation to use adoc instead of md

### LINKED ISSUE #89: Switch our docs into asciidoc format
In order for our documentation to be included in other sources we will need to switch to using asciidoc.  Here's an [example](https://github.com/eclipse/microprofile-open-api/blob/master/spec/src/main/asciidoc/microprofile-openapi-spec.adoc).

edavidj:

- Update all documentation files from markdown to asciidoc, including release README files. 
- Reconsider formatting in cases where logic can be better represented by asciidoc functionality **OR** where markdown logic cannot be effectively conveyed using asciidoc *directly* (usually indentation or custom list styles).

**Note**: Assumes that **all** documentation should be updated to use adoc. This includes the releases README.md files. If this wasn't the intention let me know.

Fixes #89


go.mod
```diff
@@ -15,7 +15,6 @@ require (
 	github.com/pkg/errors v0.8.1
 	github.com/spf13/pflag v1.0.5
 	k8s.io/api v0.17.2
-	k8s.io/apiextensions-apiserver v0.17.1
```
arthurdm, line 18: is this related to the documentation change?

edavidj: I don't believe so. I ran `go mod tidy` after rebasing onto master,  should I remove that change from the commit or leave it as it seems that go modules thinks it's not being used?

arthurdm: probably better to leave it for another PR just for historic tracking purposes.

PR #115, opened 2020-05-15
Unit test

JusteenR:

- Increased test coverage for unit testing for utils.go

### DIFF
--- a/pkg/utils/utils_test.go
+++ b/pkg/utils/utils_test.go
@@ -1,11 +1,15 @@
 package utils
 
 import (
+	"errors"
+	"fmt"
 	"os"
 	"reflect"
 	"strconv"
+	"strings"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


pkg/utils/utils_test.go
```diff
@@ -568,9 +1073,9 @@ func createAppDefinitionTags(app *appstacksv1beta1.RuntimeComponent) (map[string
 	}
 	return label, annotations
 }
-func createRuntimeComponent(n, ns string, spec appstacksv1beta1.RuntimeComponentSpec) *appstacksv1beta1.RuntimeComponent {
+func createRuntimeComponent(n, ns string, spec appstacksv1beta1.RuntimeComponentSpec, labels map[string]string, annotations map[string]string) *appstacksv1beta1.RuntimeComponent {
```
edavidj, line 1076: It might be cleaner to have this take a `metav1.ObjectMeta` as a parameter instead of four separate params.

PR #117, opened 2020-05-19
Add e2e test for kAppNav

### LINKED ISSUE #106: Add e2e test for kAppNav
- Add e2e test for kAppNav feature that:
  - Tests default state where it creates app definition
  - Disables create app definition and verifies cleanup of labels/applications
  - Configures to use existing application
  - Changes configured existing application (or update to create new one)

edavidj:

- Adds e2e testing for kAppNav to compliment existing unit tests

- Doesn't cover the case where it's connecting to an application across namespaces, going to open another issue to track progress on getting that to work for all relevant e2e tests.

Fixes #106

### DIFF
--- a/test/e2e/runtime_kappnav.go
+++ b/test/e2e/runtime_kappnav.go
@@ -1,0 +1,209 @@
+package e2e
+
+import (
+	goctx "context"
+	"errors"
+	"testing"
+	"time"
+
+	appstacksv1beta1 "github.com/application-stacks/runtime-component-operator/pkg/apis/appstacks/v1beta1"
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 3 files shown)


test/e2e/runtime_kappnav.go
```diff
@@ -0,0 +1,209 @@
...
+	if err != nil {
+		return err
+	}
+
+	// connect to existing application IN namespace
+	target := types.NamespacedName{Namespace: ns, Name: name}
+
+	err = util.UpdateApplication(f, target, func(r *appstacksv1beta1.RuntimeComponent) {
+		createApp := false
+		r.Spec.CreateAppDefinition = &createApp
```
leochr, line 161: Operator should find the Application CR matching the application name in the current namespace, even if CreateAppDefinition is not explicitly set to false. So this line should be removed.

PR #121, opened 2020-05-21
Quick fix to TLS Secret example.

InfTkm:
Added missing `.crt` after destCA. Encountered this typo while developing the E2E test.


InfTkm:
Will combine with another PR.

PR #125, opened 2020-05-26
Add node and pod affinity doc

arturdzm:


doc/user-guide.adoc
```diff
@@ -912,6 +917,107 @@ data:
...
+          matchExpressions:
+          - key: failure-domain.beta.kubernetes.io/zone
+            operator: In
+            values:
+            - zoneB
+----
+
+==== Pod Affinity and Anti-Affinity
+
+Pod affinity and anti-affinity allow you to constrain which nodes your pod is eligible to be scheduled based on labels on pods that are already running on the node rather than based on labels on node.
```
chirp1, line 984: Change to "Use pod affinity and anti-affinity to constrain which nodes your pod can be scheduled on. Add labels to pods that are already running on the node instead of putting labels on the node.".

doc/user-guide.adoc
```diff
@@ -912,6 +917,107 @@ data:
...
+
+Bellow example shows required pod affinity where pod for service A and B should be co-located in the same zone. Additionally, it is preferred not to schedule on hosts that are running service C.
```
chirp1, line 986: Tip: Avoid using "below", "above", "right", and "left"  to give the user direction for the page that they are reading.. The reason is that these words are accessibility violations because screen readers don't have the ability to give their users a sense of direction.

Are Service A and Service B in the same pod or in separate pods? Here is possible wording. I'm assuming that the two services are in different pods. maybe they aren't.
Change to "The following example shows that pod affinity is required and that the pods for service A and service B must be in the same zone. The preference for hosts that are running on service C is that they are not scheduled.".

doc/user-guide.adoc
```diff
@@ -114,6 +114,11 @@ Each `RuntimeComponent` CR must at least specify the `applicationImage` paramete
 | `route.insecureEdgeTerminationPolicy`   | HTTP traffic policy with TLS enabled. Can be one of `Allow`, `Redirect` and `None`.
 | `route.certificate`  | A YAML object representing a link:++https://cert-manager.io/docs/reference/api-docs/#cert-manager.io/v1alpha2.CertificateSpec++[Certificate].
 | `route.certificateSecretRef` | A name of a secret that already contains TLS key, certificate and CA to be used in the route. Also can contain destination CA certificate.
+| `affinity.nodeAffinity` | A YAML object representing a link:++https://v1-17.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#nodeaffinity-v1-core++[NodeAffinity].
```
leochr, line 117: Please change to "A YAML object that represents" as per Karen's suggestion. There are more instances below.

doc/user-guide.adoc
```diff
@@ -912,6 +917,110 @@ data:
...
+          labelSelector:
+            matchExpressions:
+            - key: service
+              operator: In
+              values:
+              - Service-C
+          topologyKey: kubernetes.io/hostname
+----
+
+
```
leochr, line 1023: Please add a link to Navid's examples from here.

doc/user-guide.adoc
arturdzm: I tried to follow the patter we already have for other objects, I though it is good to be consistent

PR #137, opened 2020-06-10
Expand Unit Test for Reconciler and Controller

### LINKED ISSUE #134: Expand unit test coverage for controller
- Coverage for the controller package is at roughly 40%. Try to get this at least above 80%
- Investigate organizing the unit test into independant test functions instead of one big TestReconcile, see the tests in [open-liberty-operator](https://github.com/OpenLiberty/open-liberty-operator/blob/master/pkg/controller/openliberty/openlibertyapplication_controller_test.go) for example.

InfTkm:
Expand unit test.

Fixes #134

### DIFF
--- a/pkg/controller/runtimecomponent/runtimecomponent_controller_test.go
+++ b/pkg/controller/runtimecomponent/runtimecomponent_controller_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"os"
 	"strconv"
+	"fmt"
 	"testing"
 
 	appstacksv1beta1 "github.com/application-stacks/runtime-component-operator/pkg/apis/appstacks/v1beta1"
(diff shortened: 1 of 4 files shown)


pkg/controller/runtimecomponent/runtimecomponent_controller_test.go
```diff
@@ -190,80 +206,187 @@ func TestRuntimeController(t *testing.T) {
...
+		t.Fatalf("Unable to add prometheus scheme: (%v)", err)
 	}
+
+	s.AddKnownTypes(appstacksv1beta1.SchemeGroupVersion, runtimecomponent)
+	s.AddKnownTypes(certmngrv1alpha2.SchemeGroupVersion, &certmngrv1alpha2.Certificate{})
+	s.AddKnownTypes(prometheusv1.SchemeGroupVersion, &prometheusv1.ServiceMonitor{})
 }
 
 // Helper Functions
+func makeRuntimeAndReq() (*appstacksv1beta1.RuntimeComponent, reconcile.Request) {
```
edavidj, line 383: This helper function groups two unrelated functions, which usually isn't the best practice even if it reduces duplication a bit.  Looking at it now the `reconcile.Request` returned by `createReconcileRequest` is always the same, it could instead just be a global variable initialized at the top with the others.  You could then remove this function as well as `createReconcileRequest`

Issue #64, opened 2020-03-27
Adopt service binding specification


arthurdm:
This issue / epic will track the adoption of the relevant parts of the [service binding specification](https://github.com/application-stacks/service-binding-specification).

In particular:
* Built-in support in the `RuntimeComponent` CR to gather enough information to create (under the covers) a `ServiceBindingRequest` [CR](https://github.com/application-stacks/service-binding-specification#request-service-binding).  The user won't need to specify anything pertaining to the applicationSelector / application fields.
  * Perhaps a new `bindings` top level element will be the best idea, deprecating our current usage of `service.consumes[]`

* Making our Operator `bindable`, since we expose RESTful services.  Should be as per [this definition](https://github.com/application-stacks/service-binding-specification#making-a-service-bindable)
  * This would also imply a new way of doing `service.produces` and using the proposed [binding schema](https://github.com/application-stacks/service-binding-specification#service-binding-schema)

* Mounting data as per the [spec's definition](https://github.com/application-stacks/service-binding-specification#mounting-binding-information)
  * In the future also allowing env var injection as per [this issue](https://github.com/application-stacks/service-binding-specification/issues/15)

seabaylea:
@arthurdm Would the move to `bindings` from `service.[provides|consumes]` just be a refactor?

I think it would also be useful to start to consider how we'll handle Kafka topics for µservice<->µservice message passing. In most cases that will be the same as what we're doing for REST today (the receiver just needs to be told what topic the sender is using), but there are more advanced cases where the publisher may need to create a KafkaTopic CR as well.

arthurdm:
hey @seabaylea - we're thinking the `services` parts of the `bindings` element should closely resemble the parts of the underlying `ServiceBinding` CR that the Operator will create on the user's behalf - this allows for consistent config and ease of use (i.e. can copy/paste snippets from one to the other).  

But ya - all those things are in the `consumer` side.  If there are more advanced options for the `producer` side this issue would be the place to discuss, as we can (and should) definitely take that into account.

navidsh:
Delivered. Closing.

PR #148, opened 2020-07-09
Added support to use application as a backing service

navidsh:
- Added support for making the operator act as a backing service by adding `.status.binding.name` to include the  name of a secret that contains binding information
- The operator creates a binding secret named `<cr-name>-expose-binding` and adds `host`, `port`, `protocol`, `uri` and `basePath` to the secret
- Users can override the default values by creating a second secret named `<cr-name>-expose-binding-override` to override or add any k/v to the binding secret
- Any changes to the binding secret or the override secret causes a reconcile on the CR so that secrets get the latest values
- Users can expose the binding secret by explicitly setting `.spec.bindings.expose.enabled` to `true` in the CR

Doc to be delivered in a separate PR

Fixes #


navidsh:
To make SBO recognize the binding secret, the following should be added under `statusDescriptors` in CSV:
```yaml
      - description: The name of the secret object that stores the service binding information.
        displayName: Service Binding Secret
        path: binding.name
        x-descriptors:
        - urn:alm:descriptor:io.kubernetes:Secret
        - binding:env:object:secret
```

navidsh:
Fixes #64

PR #157, opened 2020-08-13
Add support for generic day-2 operations

leochr:

- Add support for generic day-2 operations


pkg/apis/appstacks/v1beta1/operations.go
```diff
@@ -0,0 +1,85 @@
...
+	LastUpdateTime     metav1.Time                  `json:"lastUpdateTime,omitempty"`
+	Reason             string                       `json:"reason,omitempty"`
+	Message            string                       `json:"message,omitempty"`
+	Status             corev1.ConditionStatus       `json:"status,omitempty"`
+	Type               OperationStatusConditionType `json:"type,omitempty"`
+}
+
+// OperatedResource ...
+// +k8s:openapi-gen=true
+type OperatedResource struct {
```
navidsh, line 23: I couldn't find any usage for this struct. Why do we need it again?

pkg/controller/runtimeoperation/runtimeoperation_controller.go
```diff
@@ -0,0 +1,189 @@
...
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return isClusterWide || watchNamespacesMap[e.Meta.GetNamespace()]
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
+			return isClusterWide || watchNamespacesMap[e.Meta.GetNamespace()]
+		},
+	}
+
+	// Watch for changes to primary resource RuntimeOperation
+	err = c.Watch(&source.Kind{Type: &appstacksv1beta1.RuntimeOperation{}}, &handler.EnqueueRequestForObject{}, pred)
```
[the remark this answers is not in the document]
leochr: Yeah, watch is needed to emit that event and to reconcile when the operation is not started (i.e. incorrect pod name).

pkg/apis/appstacks/v1beta1/operations.go
leochr: It's not used. Removed it.

pkg/controller/runtimeoperation/runtimeoperation_controller.go
```diff
@@ -0,0 +1,189 @@
...
+		log.Info(message)
+		r.recorder.Event(instance, "Warning", "ProcessingError", message)
+		return reconcile.Result{}, err
+	}
+
+	//check if Pod exists and running
+	pod := &corev1.Pod{}
+
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: instance.Spec.PodName, Namespace: request.Namespace}, pod)
+	if err != nil || pod.Status.Phase != corev1.PodRunning {
```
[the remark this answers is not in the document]
leochr: Good point. Made code changes, so the operator reconciles when the operation fails to start due to incorrect pod or container name or pod is not in running state.

doc/user-guide.adoc
```diff
@@ -1034,6 +1034,44 @@ spec:
...
+  # Command to execute. Not executed within a shell
+  command:
+    - /bin/sh
+    - '-c'
+    - echo "Hello" > /tmp/runtime-operation.log
+----
+
+You can check the status of a runtime operation using the `status` field inside the CR YAML. You can also run the command `oc get runtimeop -o wide` to see the status of all operations in the current namespace.
+
+NOTE: The `RuntimeOperation` CR must be created in the same namespace as the `RuntimeComponent` to operate on. Once the `RuntimeOperation` has started, the CR can not be re-used for more operations. A new CR needs to be created for each day-2 operation. The operator can process only one `RuntimeOperation` at a given time. Long running commands can queue up the other runtime operations.
```
chirp1: A noun is needed after `RuntimeComponent`. I'm not sure what "to operate on" at the end of the sentence means. Is "to operate on" needed? How about changing "The `RuntimeOperation` CR must be created in the same namespace as the `RuntimeComponent` to operate on." to something like "The `RuntimeOperation` CR must be created in the same namespace as the `RuntimeComponent` instance.".

Change "Once the `RuntimeOperation` has started, the CR can not be re-used for more operations." to "**After** the `RuntimeOperation` **CR** **starts**, the CR **cannot** be reused for more operations.".

Change "only one `RuntimeOperation` at a given time." to "only one `RuntimeOperation` instance at a time.".

I'm not sure what "Long running commands can queue up the other runtime operations." means. Does it mean that other runtim operations have to wait because the commands take so long to run? If so, how about "Long running commands can cause other runtime operations to wait before they start."? If not, reword to express the intended meaning.

PR #159, opened 2020-08-25
remove cert tests for now

edavidj:

- Disable the cert manager test until it can be fixed


leochr:
The test cluster had a newer version of cert-manager that's not compatible/supported. Installing the older version hopefully resolves the test failure. Closing this PR.

Issue #178, opened 2021-05-17
Convert tests to run on new framework


leochr:
Existing tests are not compatible with operator-sdk 1.x. Convert existing tests to one of the new testing framework (kuttl, scorecard, env-test). Tests related to kAppNav integration and servicebinding doesn't have to be ported, but we still want to keep generating Application definition)

e2e tests at https://github.com/application-stacks/runtime-component-operator/tree/master/test/e2e : 
- [x] runtime_basic.go - Jason
- [x] runtime_pullpolicy.go - Bradley
- [x] runtime_probe.go - Iain
- [x] runtime_autoscaling.go - Iain
- [x] runtime_imagestream.go - Iain
- [x] runtime_knative.go - Iain
- [x] runtime_storage.go -Iain
- [x] runtime_servicemonitor.go - Iain
- [ ] runtime_test.go - I don't believe this is actually a test, but rather the entry point for e2e tests (of all framework?)
- [x] ~runtime_kappnav.go~ - kAppNav integration will be removed in v1, but operator will continue to generate Application definition. Operator will need to be enhanced to add this support. So we don't need to convert this now.
- [x] ~runtime_certmanager.go~ - Cert manager support will change in v1, so we don't need to convert this now.
- [ ] ~runtime_servicebinding.go~ - Service binding support will change in v1, so we don't need to convert this now. - handled via https://github.com/application-stacks/runtime-component-operator/issues/239

Unit tests: handled via https://github.com/application-stacks/runtime-component-operator/issues/237

yongja79:
The script to run the tests in travis will also need to be updated [e2e.sh](https://github.com/application-stacks/runtime-component-operator/blob/master/scripts/e2e.sh)

PR #182, opened 2021-06-25
Kuttl tests

idlewis:
Adds some tests based on the kuttl test framework. These will eventually replace the existing e2e tests, which use the older operator-sdk test framework which is no longer supported

This is part of the work for issue #178


idlewis:
Closing this in favour of PR #186 which is off an updated branch

PR #186, opened 2021-07-29
Add e2e tests based on kuttl

idlewis:
Adds some tests based on the kuttl test framework. These will eventually replace the existing e2e tests, which use the older operator-sdk test framework which is no longer supported

This is part of the work for issue #178

### DIFF
--- a/bundle/tests/scorecard/kuttl/knative2/00-runtime-knative.yaml
+++ b/bundle/tests/scorecard/kuttl/knative2/00-runtime-knative.yaml
@@ -1,0 +1,10 @@
+apiVersion: app.stacks/v1beta1
+kind: RuntimeComponent
+metadata:
+  name: knative2-rc
+spec:
+  # Add fields here
+  applicationImage: navidsh/demo-day
+  replicas: 1
+  createKnativeService: true
+
--- a/bundle/tests/scorecard/kuttl/monitor/02-assert.yaml
+++ b/bundle/tests/scorecard/kuttl/monitor/02-assert.yaml
@@ -1,0 +1,46 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+timeout: 60
+---
+apiVersion: apps/v1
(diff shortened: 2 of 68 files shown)


bundle/tests/scorecard/kuttl/knative2/00-runtime-knative.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: app.stacks/v1beta1
+kind: RuntimeComponent
+metadata:
+  name: knative2-rc
+spec:
+  # Add fields here
+  applicationImage: navidsh/demo-day
+  replicas: 1
+  createKnativeService: true
```
leochr, line 9: When knative is enabled, Deployment shouldn't exist, so we should an errors.yaml with Deployment

bundle/tests/scorecard/kuttl/monitor/02-assert.yaml
```diff
@@ -0,0 +1,46 @@
...
+    tlsConfig:
+      insecureSkipVerify: true
+    basicAuth:
+      password:
+        key: password
+      username:
+        key: username
+---
+apiVersion: app.stacks/v1beta1
+kind: RuntimeComponent
```
leochr, line 40: What is the purpose of this RuntimeComponent snippet in assert?

idlewis: Not sure, that looks like a mistake. It is only asserting on the spec, which should never be incorrect, so I can't see that there is a point. I will take that out.

bundle/tests/scorecard/kuttl/knative2/00-runtime-knative.yaml
idlewis: I'll add that.

leochr (approved):
@idlewis thanks for responding to the review comments quickly

PR #193, opened 2021-08-17
Issue #173: Adding update strategy configuration support

### LINKED ISSUE #173: Allow configuration of Deployment strategy
We need to surface a way to configure the [Deployment.spec.strategy](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#update-strategies) and [StatefulSet.spec.updateStrategy](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#update-strategies) that we create.

halim-lee:
- Users can configure the update strategy for deployment and statefulset

- [x] User guide
- [ ] `CHANGELOG.md`

Fixes #173


common/types.go
```diff
@@ -187,4 +188,6 @@ type BaseComponent interface {
 	GetRoute() BaseComponentRoute
 	GetBindings() BaseComponentBindings
 	GetAffinity() BaseComponentAffinity
+	GetDeploymentStrategy() *appsv1.DeploymentStrategy
```
leochr, line 191: We'll likely have more parameters under deployment and statefulSet in the future. It'll be good to create these update strategies under another level - similar to how GetService() BaseComponentService is setup. Then BaseComponentService has various methods (i.e. GetPort() int32)

leochr (commented):
@halim-lee Thanks for the PR. Looks good. Added a comment about the parent elements deployment/statefulset.

utils/utils.go
```diff
@@ -47,6 +47,11 @@ func CustomizeDeployment(deploy *appsv1.Deployment, ba common.BaseComponent) {
 		}
 	}
 
+	dp := ba.GetDeployment()
+	if dp.GetDeploymentUpdateStrategy() != nil {
```
leochr, line 51: wondering if a nil check on `dp` is needed since `spec.deployment` might not be set in the CR.

leochr (commented):
@halim-lee Thanks for the updates. Looks good. Just one comment/clarification about a nil check.

leochr (approved):
@halim-lee Looks good. Thanks.

PR #194, opened 2021-08-19
Kuttl tests4

BradleyMayo:
Adds runtime pull policy tests
- 
- 

- [ ] User guide
- [ ] `CHANGELOG.md`

Fixes #

### DIFF
--- a/bundle/tests/scorecard/kuttl/runtime_pullpolicy/00-assert.yaml
+++ b/bundle/tests/scorecard/kuttl/runtime_pullpolicy/00-assert.yaml
@@ -1,0 +1,16 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+timeout: 60
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: example-runtime-component
+status:
+  readyReplicas: 1
+---
+apiVersion: v1
+kind: Pod
+spec:   
+  containers:
+  - imagePullPolicy: "Always"
--- a/bundle/tests/scorecard/kuttl/runtime_pullpolicy/01-assert.yaml
+++ b/bundle/tests/scorecard/kuttl/runtime_pullpolicy/01-assert.yaml
@@ -1,0 +1,16 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+timeout: 60
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: example-runtime-component
+status:
+  readyReplicas: 1
+---
+apiVersion: v1
+kind: Pod
+spec:   
+  containers:
+  - imagePullPolicy: "ifNotPresent"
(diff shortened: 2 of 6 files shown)


bundle/tests/scorecard/kuttl/runtime_pullpolicy/02-never.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: app.stacks/v1beta1
+kind: RuntimeComponent
+metadata:
+  name: example-runtime-component
+spec:
+  applicationImage: navidsh/demo-day
+  replicas: 1
+  pullPolicy: "Never"
```
leochr, line 8: Please add a test to verify that removing `spec.pullPolicy` from the `RuntimeComponent` CR, results in Deployment/Pod being created with the default pull policy values (in this case the image is referring to _latest_ tag implicitly, hence pull policy the k8s resource(s) should be _Always_ I believe).

bundle/tests/scorecard/kuttl/runtime_pullpolicy/00-assert.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+timeout: 60
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: example-runtime-component
```
leochr, line 8: It'll be good to add a check for Deployments as well - that it has the right imagePullPolicy

bundle/tests/scorecard/kuttl/runtime_pullpolicy/01-assert.yaml
```diff
@@ -0,0 +1,16 @@
...
+metadata:
+  name: example-runtime-component
+status:
+  readyReplicas: 1
+---
+apiVersion: v1
+kind: Pod
+spec:   
+  containers:
+  - imagePullPolicy: "ifNotPresent"
```
leochr, line 16: Not sure if capitalization matters, but should be _IfNotPresent_

leochr (commented):
@BradleyMayo Thanks for the PR. Added few comments.

bundle/tests/scorecard/kuttl/runtime_pullpolicy/01-assert.yaml
idlewis: It does matter. The test fails in its current state

bundle/tests/scorecard/kuttl/runtime_pullpolicy/03-default.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: app.stacks/v1beta1
+kind: RuntimeComponent
+metadata:
+  name: example-runtime-component
+spec:
+  applicationImage: navidsh/demo-day
+  replicas: 1
```
idlewis, line 7: This test step fails. The new spec is merged with the previous spec, so you need to explicitly remove the configured pullPolicy. I think you can either add
```pullPolicy: ```
or
```pullPolicy: null```
to the end of this test file

idlewis (approved):

PR #198, opened 2021-08-20
Add kuttl test for persistent storage

idlewis:
Adding this test as part of issue #178.
It was missed out from the original PR

### DIFF
--- a/bundle/tests/scorecard/kuttl/persistent-storage/00-persistent-storage-rc.yaml
+++ b/bundle/tests/scorecard/kuttl/persistent-storage/00-persistent-storage-rc.yaml
@@ -1,0 +1,19 @@
+apiVersion: app.stacks/v1beta1
+kind: RuntimeComponent
+metadata:
+  name: persistent-rc
+spec:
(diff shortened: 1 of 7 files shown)


bundle/tests/scorecard/kuttl/persistent-storage/00-persistent-storage-rc.yaml
```diff
@@ -0,0 +1,19 @@
...
+kind: RuntimeComponent
+metadata:
+  name: persistent-rc
+spec:
+  applicationImage: navidsh/demo-day
+  replicas: 1
+  volumeMounts:
+    - name: pvc
+      mountPath: /data
+  storage:
```
leochr, line 11: Specifying `spec.storage.mountPath: /data` should have similar result. The operator will specify the volume mount automatically. It'll be good to add a test case for that scenario.

leochr (approved):
@idlewis Thanks Iain. Looks good. Added a comment for an additional test to cover related persistence scenario. That can be done via a separate PR.

PR #204, opened 2021-09-07
Adding StartupProbe and UpdateStrategy Tests

### LINKED ISSUE #201: Add tests for new functions
- [ ] startup probe support
- [ ] update strategy for Deployment and StatefulSet

halim-lee:

- Tests for Startup Probe and Deployment & StatefulSet Update Strategies

- [ ] User guide
- [ ] `CHANGELOG.md`

Fixes #201


idlewis (commented):
For the startup probe test, is it worth modifying step 1 of the test to change the times of the startup probe? that would match what the liveliness/readiness sections of the test do

bundle/tests/scorecard/kuttl/deployment-strategy/01-update-application.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: app.stacks/v1beta1
```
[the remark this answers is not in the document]
halim-lee: > For the startup probe test, is it worth modifying step 1 of the test to change the times of the startup probe? that would match what the liveliness/readiness sections of the test do

Yes, I will add modification to startup probe times in test 1.

> Is this whole step necessary? The only change is to the application image, and the rollout itself is handled by kube, so it feels a bit as if we are testing the kube rollout strategy. Unless I've missed something, I think it would probably be better to keep the test short and remove this step
But it might be worth adding a step to explicitly set the strategy to recreate.

I will remove this step and keep test 3 for specifying strategy to Recreate or OnDelete.

idlewis (approved):

bundle/tests/scorecard/kuttl/probe/02-runtime-remove-probe.yaml
```diff
@@ -6,3 +6,4 @@ spec:
   # Add fields here
   livenessProbe: null
   readinessProbe: null
+  startupProbe: null
```
leochr, line 9: For this scenario, we should check that startupProbe is no longer defined by adding it to 02-errors.yaml

bundle/tests/scorecard/kuttl/deployment-strategy/01-remove-rolling-update.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: RuntimeComponent
+metadata:
+  name: deployment-strategy-rc
+spec:
+  # Add fields here
+  replicas: 3
+  deployment:
+    updateStrategy:
+      rollingUpdate: null
+      type: Recreate
```
leochr, line 11: Following this check, we should remove updateStrategy altogether and make sure Deployment no longer has the Recreate type (Deployment will default to RollingUpdate). Same for StatefulSet.

leochr (commented):
@halim-lee Looks good. Thanks. Added some minor comments to further enhance the tests.

bundle/tests/scorecard/kuttl/deployment-strategy/04-remove-deployment.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: app.stacks/v1beta1
+kind: RuntimeComponent
+metadata:
+  name: deployment-strategy-rc
+spec:
+  # Add fields here
+  replicas: 3
+  deployment:
```
leochr, line 8: @halim-lee Tanks for updating the tests. The flow looks good. Wondering whether `deployment` field can be removed altogether here? or does kuttl require it to be set this way?

If it can be removed, please do the same for updateStrategy field. Thanks.

halim-lee: kuttl test updates what is specified in the yaml file. If I remove deployment field, it wouldn't know deployment should be removed - so I made the deployment field empty.

leochr (approved):

Issue #205, opened 2021-09-08
PVC not getting created through volumeClaimTemplate


halim-lee:
### What did you do?

When I try to create persistent volume for a statefulset through volumeClaimTemplate, PVC does not get created.

### What did you expect to see?

PVC should be created with declared fields in volumeClaimTemplate.

### What did you see instead?

```
spec:
  storage:
    volumeClaimTemplate:
      metadata: {}
      spec:
        accessModes:
          - ReadWriteOnce
        resources:
          requests:
            storage: 50Mi
        storageClassName: local-storage
      status: {} 
```
metadata and status fields are empty - which looks like the cause of PVC not getting created.

### Environment

OpenShift 4.7.13, Kubernetes 1.20

halim-lee:
kubernetes-sigs/controller-tools#557

The bug was from controller-tools, which cleared all fields under ObjectMeta when preserveUnknownFields is set to False. It is fixed in v0.6.0 through using generateEmbeddedObjectMeta option.

PR #206, opened 2021-09-14
Issue #205: Volume Claim Template fix

halim-lee:

- 
- 

- [ ] User guide
- [ ] `CHANGELOG.md`

Fixes #


halim-lee:
https://github.com/kubernetes-sigs/controller-tools/pull/557

The bug was from controller-tools, which cleared all fields under `ObjectMeta` when `preserveUnknownFields` is set to False. It is fixed in v0.6.0 through using `generateEmbeddedObjectMeta` option.

arturdzm (approved):

leochr:
@halim-lee Thanks for digging into this and finding out the root cause. 

The release note of controller-tools v0.6.0, states the following:
_Update Kubernetes 1.21.1 and Go 1.16_  

Would this affect our support on OpenShift 4.6 (long term supported version) which uses Kubernetes 1.19? I want to make sure there will be no compatibility issues. @arturdzm do you know?

arturdzm:
As far as  I know kubernetes API supposed to be compatible with +- 2 version of kubernetes. So tools like kubectl
can be 2 versions below or above server version.

PR #207, opened 2021-09-17
Issue #195: StatefulSet creation without PVC for each instance

### LINKED ISSUE #195: Allow StatefulSet creation without requiring PVC for each instance
The operator's current behaviour is to create StatefulSet (instead of a Deployment) when `spec.storage.*` is set. This would require each instance/pod to have its own PVC. Sometimes StatefulSet is needed for the unique name each Pod gets. Persistence is handled via db/storage that's shared by all pods. To support this scenario, we should create StatefulSet when `spec.statefulSet` is not empty. Also, move storage config from `spec.storage.*` to `spec.statefulSet.storage.*`

halim-lee:
- To create statefulset without requiring PVC for each instance
- storage field is moved under statefulSet field

- [x] User guide
- [ ] `CHANGELOG.md`

Fixes #195

### DIFF
--- a/bundle/tests/scorecard/kuttl/out.txt
+++ b/bundle/tests/scorecard/kuttl/out.txt
@@ -1,0 +1,21 @@
+=== RUN   kuttl
+    harness.go:457: starting setup
+    harness.go:248: running tests using configured kubeconfig.
+    harness.go:285: Successful connection to cluster at: https://9.30.235.22:6443
+    harness.go:353: running tests
+    harness.go:74: going to run test suite with timeout of 300 seconds for each step
+    harness.go:365: testsuite: ./ has 13 tests
+=== RUN   kuttl/harness
+=== RUN   kuttl/harness/persistent-storage
+=== PAUSE kuttl/harness/persistent-storage
+=== RUN   kuttl/harness/teststorage
+=== PAUSE kuttl/harness/teststorage
+=== CONT  kuttl/harness/persistent-storage
+    logger.go:42: 14:49:53 | persistent-storage | Ignoring README.txt as it does not match file name regexp: ^(\d+)-(?:[^\.]+)(?:\.yaml)?$
+    logger.go:42: 14:49:53 | persistent-storage | Skipping creation of user-supplied namespace: operator-test
+    logger.go:42: 14:49:53 | persistent-storage/0-persistent-storage-rc | starting test step 0-persistent-storage-rc
+    logger.go:42: 14:49:57 | persistent-storage/0-persistent-storage-rc | RuntimeComponent:operator-test/persistent-rc created
+=== CONT  kuttl
+    harness.go:508: cleaning up
+    harness.go:563: removing temp folder: ""
+    harness.go:444: failed with interrupt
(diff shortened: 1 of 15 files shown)


utils/utils.go
```diff
@@ -690,13 +691,14 @@ func CustomizeHPA(hpa *autoscalingv1.HorizontalPodAutoscaler, ba common.BaseComp
...
 func Validate(ba common.BaseComponent) (bool, error) {
 	// Storage validation
-	if ba.GetStorage() != nil {
-		if ba.GetStorage().GetVolumeClaimTemplate() == nil {
-			if ba.GetStorage().GetSize() == "" {
+	ss := ba.GetStatefulSet()
+	if ss != nil && ss.GetStorage() != nil {
+		if ss.GetStorage().GetVolumeClaimTemplate() == nil {
+			if ss.GetStorage().GetSize() == "" {
 				return false, fmt.Errorf("validation failed: " + requiredFieldMessage("spec.storage.size"))
```
leochr, line 698: It should be spec.statefulSet.storage.size

bundle/tests/scorecard/kuttl/out.txt
```diff
@@ -0,0 +1,21 @@
+=== RUN   kuttl
```
leochr, line 1: I believe this is produced by running kuttl locally. Probably should be added to gitignore. Not sure about the kubeconfig above

bundle/tests/scorecard/kuttl/persistent-storage/00-persistent-storage-rc.yaml
```diff
@@ -8,12 +8,13 @@ spec:
   volumeMounts:
```
leochr, line 8: we should add a test to make sure StatefulSet is created and not Deployment when `statefulSet: {}` is specified in CR. Fine with doing that via a separate PR.

utils/utils.go
```diff
@@ -680,7 +681,7 @@ func CustomizeHPA(hpa *autoscalingv1.HorizontalPodAutoscaler, ba common.BaseComp
 	hpa.Spec.ScaleTargetRef.Name = obj.GetName()
 	hpa.Spec.ScaleTargetRef.APIVersion = "apps/v1"
 
-	if ba.GetStorage() != nil {
+	if ba.GetStatefulSet() != nil && ba.GetStatefulSet().GetStorage() != nil {
```
leochr, line 684: This should check only for `ba.GetStatefulSet() != nil` - because StatefulSet will be created for that scenario (whether storage is configured or not).

controllers/runtimecomponent_controller.go
```diff
@@ -324,7 +324,10 @@ func (r *RuntimeComponentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		return r.ManageError(err, common.StatusConditionTypeReconciled, instance)
 	}
 
-	if instance.Spec.Storage != nil {
+	if instance.Spec.StatefulSet != nil {
+		l := reqLogger.WithValues("instance.Spec.StatefulSet!!!!!", instance.Spec.StatefulSet)
```
leochr, line 328: These logging statements can probably be removed.

doc/user-guide.adoc
```diff
@@ -37,9 +37,10 @@ spec:
     type: ClusterIP
```
leochr, line 37: The user-guide.adoc in master branch is used by current users of beta operator releases, but these 1.0 doc changes wouldn't apply to their version. Please clone the content to user-guide-v1.adoc and add these there. We should also remove the previous set of doc changes from user-guide.adoc (i.e. startupProbe)

leochr (commented):
@halim-lee Thanks for the PR! Looks good for the most part. Added some comments.

PR #209, opened 2021-09-21
UpdateStrategy to RollingUpdate when nil

### LINKED ISSUE #208: UpdateStrategy not getting cleared
### What did you do?
When I set the update strategy to "OnDelete" for statefulSet or "Recreate" for deployment, then set them to null, the update strategy remains as OnDelete or Recreate.

```
  statefulSet:
    updateStrategy:
      type: OnDelete
    storage:
      size: 1Gi
      mountPath: "/logs"
```

```
  statefulSet:
    updateStrategy:
    storage:
      size: 1Gi
      mountPath: "/logs"
```

When update strategy field is null, it should set back to RollingUpdate.

halim-lee:
- Deployment and StatefulSet update strategy should fall back to RollingUpdate when it it set to nil.

- [ ] User guide
- [ ] `CHANGELOG.md`

Fixes #208

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -48,14 +48,11 @@ func CustomizeDeployment(deploy *appsv1.Deployment, ba common.BaseComponent) {
 	}
 
 	dp := ba.GetDeployment()
-	if dp != nil {
-		if dp.GetDeploymentUpdateStrategy() != nil {
-			deploy.Spec.Strategy = *dp.GetDeploymentUpdateStrategy()
-		} else {
-			deploy.Spec.Strategy = appsv1.DeploymentStrategy{Type: appsv1.RollingUpdateDeploymentStrategyType}
-		}
+	if dp != nil && dp.GetDeploymentUpdateStrategy() != nil {
+		deploy.Spec.Strategy = *dp.GetDeploymentUpdateStrategy()
+	} else {
+		deploy.Spec.Strategy = appsv1.DeploymentStrategy{Type: appsv1.RollingUpdateDeploymentStrategyType}
 	}
-
 }
 
 // CustomizeStatefulSet ...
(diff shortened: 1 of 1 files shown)


utils/utils.go
```diff
@@ -48,8 +48,12 @@ func CustomizeDeployment(deploy *appsv1.Deployment, ba common.BaseComponent) {
 	}
 
 	dp := ba.GetDeployment()
-	if dp != nil && dp.GetDeploymentUpdateStrategy() != nil {
-		deploy.Spec.Strategy = *dp.GetDeploymentUpdateStrategy()
+	if dp != nil {
+		if dp.GetDeploymentUpdateStrategy() != nil {
+			deploy.Spec.Strategy = *dp.GetDeploymentUpdateStrategy()
+		} else {
+			deploy.Spec.Strategy = appsv1.DeploymentStrategy{Type: appsv1.RollingUpdateDeploymentStrategyType}
```
leochr, line 55: The user could remove `spec.deployment` altogether. We would still have a Deployment, but with old/removed strategy. We should set the default strategy above the `dp != nil` check. It might be a good idea to add a test for this scenario as well. Make sure to specify a non-default update strategy beforehand, so we can confirm that it in fact went back to default. Sorry to pile on more requirements :) 

This is not an issue for StatefulSet, since user has to explicitly specify `spec.statefulSet` in order to create StatefulSet.

leochr (commented):
@halim-lee Thanks for the PR. Added a comment

halim-lee:
The code is updated as setting updateStrategy as RollingUpdate if deployment or updateStrategy is nil.

The test case is added in https://github.com/application-stacks/runtime-component-operator/pull/204! :)
(0. Checking RollingUpdate, 1. Checking Recreate, 2. Checking when updateStrategy is nil - falling back to RollingUpdate, 3. Re-setting to Recreate,  4. Checking when deployment is nil - falling back to RollingUpdate)

leochr (approved):
@halim-lee Looks good. Thanks

Issue #213, opened 2021-09-29
Add test cases for user deleting configuration from spec


leochr:
When user delete the configuration, the operator should remove the relevant configuration and restore relevant k8s resources (Deployment/StatefulSet, Service, Route, PVC) to the default state.

idlewis:
Most of the existing kuttl E2E tests already have a final step which deletes any modified config, and then checks that state returns to the default. I've opened PR #214 to add similar steps the the two tests which didn't, the basic test and the service monitor test

PR #214, opened 2021-10-01
Test improvements

idlewis:
Add a couple of new test steps that check if resources revert
to the default state when modifications are removed

This PR addresses issue #213


leochr (approved):
@idlewis Looks good. Thanks.

Issue #216, opened 2021-10-04
Add tests for the certificates support


leochr:
Certificates can be specified on Service and Route.

Old test to use as reference: https://github.com/application-stacks/runtime-component-operator/blob/v0.7.1/test/e2e/runtime_certmanager.go 

We've removed the embedded certificate support (removed spec.service.certificate and spec.route.certificate). So just need tests for these : spec.service.certificateSecretRef  and spec.route.certificateSecretRef)

PR #217, opened 2021-10-06
Add tests for route/service with secret ref

idlewis:
secret ref references a dummy certificate.

This addresses issue #216


leochr (approved):
@idlewis Looks good. Thanks

PR #219, opened 2021-10-11
[WIP] Enable e2e test

### LINKED ISSUE #218: Create Pipeline for Delivery
Will create a brand new project for every build with a unique name, would deploy the secret, deploy the bundle, then run the tests

Build image => Bundle Build => Pushing to cluster => Operator SDK is used to create the catalog => Install Catalog

## Starting Point:
Start with pushing the images(unique image of operator and bundle) into the custom project
Then start with bundle command that will install the operator
Can clean the cluster by deleting the project

## Todos:

- [x] Use pull requests to kick off travis/pipeline
- [x] Deploy image and bundle into OCP environment
    - [x] Need login to the docker hub registry in the cluster(using OC) - To avoid rate limits
    - [x] Make file will create image (Make break builds if we have changes, but only master works) - We can test changes as long we put them in a branch since branch can run on travis
    - [x] Need login to the cluster(using OC) to deploy
    - [x] Get deploy command from Artur 
- [x] Kuttl tests need to be executed after deploying the bundle
- [x] Can clean the cluster by deleting the project

## Stretch Goal:

- [ ] Limiting concurrent builds
- [ ] Contingency for the cloud environment to go away

hibell:
Enable e2e testing of the operator.

No.

Fixes #218.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -16,3 +16,4 @@ Currently the projects that are importing this library are:
 
 * Appsody Operator: https://github.com/appsody/appsody-operator
 * Open Liberty Operator: https://github.com/OpenLiberty/open-liberty-operator
+


arturdzm:
One suggestion would be to put the e2e into a separate step after verify-image build and not part of the arch matrix, the reason is that tests might fail but other arch could push  respective arch images regardless.

scripts/e2e.sh
```diff
@@ -1,59 +1,132 @@
...
+readonly usage="Usage: e2e.sh -u <docker-username> -p <docker-password> --cluster-url <url> --cluster-token <token> --registry-name <name> --registry-namespace <namespace>"
 readonly SERVICE_ACCOUNT="travis-tests"
 
-# login_cluster : Download oc cli and use it to log into our persistent cluster
-login_cluster(){
+# setup_env: Download oc cli, log into our persistent cluster, and create a test project
+setup_env() {
+    echo "****** Installing OC CLI..."
     # Install kubectl and oc
     curl -L https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable-4.3/openshift-client-linux.tar.gz | tar xvz
```
leochr, line 10: We should update to at least 4.6.0 version (long supported version): https://mirror.openshift.com/pub/openshift-v4/clients/ocp/4.6.0/openshift-client-linux.tar.gz

PR #222, opened 2021-10-14
Issue OLO#236: TLS secret timing issue fix

halim-lee:
- There is a timing issue with TLS secret creation. TLS secret is required to pull image from OSCP image registry and it is often not created when the operator tries to pull the image.
- TLS secret will now be created before creating deployment/statefulset

- [ ] User guide
- [ ] `CHANGELOG.md`


leochr:
Recent fix to check for service account's pull secret should resolve this issue.

PR #240, opened 2021-11-02
OLO Issue #222: Adding versions to resources in CSV

halim-lee:

- Version tag for resources in CSV is required

- [ ] User guide
- [ ] `CHANGELOG.md`

Fixes https://github.com/OpenLiberty/open-liberty-operator/issues/222 for RCO.


config/manifests/bases/runtime-component.clusterserviceversion.yaml
```diff
@@ -15,290 +15,31 @@ spec:
   customresourcedefinitions:
     owned:
     - description: RuntimeComponent is the Schema for the runtimecomponents API.
```
leochr, line 17: Wondering why these are being removed by the make commands on your machine. Are you using operator-sdk 1.6.4? Did you run `make generate && make manifests && make bundle`?

leochr (commented):
@halim-lee Thanks. Looks good. Added a question about the x-descriptors being removed from the config/base files.

config/manifests/bases/runtime-component.clusterserviceversion.yaml
leochr: Discussed this with Melissa, and this is duplicate content (resulted from v1beta1 to v1beta2 change), so removing is the right step.

PR #242, opened 2021-11-03
Add test for conflicting annotations

idlewis:
This is a second PR to address issues from issues #213.
Currently, annotations are not removed from the resources when they are removed from the spec, as the operator can't tell if the annotations were added by itself or by openshift, and so doesn't know which annotations to remove.
So this PR is just adding the extra requested tests from the comment in PR https://github.com/application-stacks/runtime-component-operator/pull/211#pullrequestreview-780449333


leochr (approved):
@idlewis Looks good. Thanks

Issue #177, opened 2021-05-12
Leverage kustomize or helm to simplify manifest-based deployment


ron1:
## Feature Request
Leverage kustomize or helm to simplify manifest-based deployments from the deploy/releases directory. Using tools like curl and sed to deploy the operator breaks the yaml-based declarative approach endorsed by gitops best practices.

leochr:
@idlewis please add the kustomize files for the 0.8.0 release at deploy/releases/0.8.0/kustomize

PR #256, opened 2021-11-11
Kustomise files for daily

idlewis:
start to deal with #177

### DIFF
--- a/deploy/kustomize/daily/README.adoc
+++ b/deploy/kustomize/daily/README.adoc
@@ -1,0 +1,43 @@
+= Installing the Runtime Component Operator using kustomize
+
+This directory contains kustomize files that can be used to install the operator
+in your cluster in various different configurations, and also some example overlays
+which show how the installation can be customized
+
+== base
+The simplest configuration will install the operator into the default namespace, and the operator
+will watch for RuntimeComponent instances only in its own namespace. To install, run:
+`kubectl apply -k base`
+and to uninstall, run:
+`kubectl delete -k base`
+
+== examples/watch-own-namespace
+This example overlay demonstrates how to modify the base configuration to install/watch a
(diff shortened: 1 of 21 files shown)


leochr (commented):
@idlewis Thanks. Looks good. Just want to confirm the purpose of _all-namespaces2_. Is it to specify a different namespace? Wondering whether changing `namespace: default` in all-namespaces/kustomization.yaml would achieve the same

idlewis:
@leochr Yes, the purpose of the all-namespaces2 is to specify a different a namespace, and yes this could be done in all-namespaces. However, the idea was that, because all-namespaces adds in a new resource file (the cluster roles yaml) I was expecting that users would want to treat all-namespaces as a base to build on, and probably would prefer to leave it untouched, rather than having to edit it. So all-namespaces2 is an example of how a user could build on all-namespaces without having to modify it.
If you think it is better to combine the two, I'm happy to do that.
I also will add a readme to explain it either way, once you are happy with the content.

leochr:
Thanks @idlewis . It'll be good to separate the folders that customers should modify from the folders/files we provide (i.e. move _all-namespace2_ and _test-namespace_ under `examples`)

Suggest the following names for the folders:
watch-own-namespace
watch-another-namespace
watch-all-namespaces

leochr (commented):
@idlewis Thanks for the update. If possible, it'll be good to have the examples directly under `daily` folder. This way that's the only folder customers would have to edit and also avoid accidentally modifying other folders directly under overlays. 

Maybe you are planning to add, it'll be good to have an example for watch-another-namespace.

idlewis:
@leochr I think I've addressed all the review comments. Hopefully should be ready for review/merge.

leochr (approved):
@idlewis Looks good. Thanks.

PR #257, opened 2021-11-15
Day 2 Operations kuttl test

### LINKED ISSUE #251: Add tests for RuntimeOperation
Execute a day-2 operation on a Pod using RuntimeOperation CR.

https://github.com/application-stacks/runtime-component-operator/blob/master/doc/user-guide-v1beta2.adoc#day-2-operations

The pod name is always random for a Deployment. Use `statefulSet: {}` to specify a sticky name for the Pod and use that in the test. 

This functionality is a generic version of OLO's OpenLibertyTrace. So the tests from OLO can be used as reference: 
https://github.com/OpenLiberty/open-liberty-operator/tree/master/bundle/tests/scorecard/kuttl/trace

yongja79:

- Adds kuttl tests for Day 2 Operations

Fixes #251

### DIFF
--- a/bundle/tests/scorecard/kuttl/day2operation/01-assert.yaml
+++ b/bundle/tests/scorecard/kuttl/day2operation/01-assert.yaml
@@ -1,0 +1,10 @@
+apiVersion: rc.app.stacks/v1beta2
+kind: RuntimeOperation
+metadata:
+  name: example-runtime-operation-0
+spec:
+  podName: day2-operation-rc-0
+  containerName: app
+status:
+  conditions:
+    - status: 'True'
--- a/bundle/tests/scorecard/kuttl/day2operation/01-assert.yaml
+++ b/bundle/tests/scorecard/kuttl/day2operation/01-assert.yaml
@@ -7,4 +7,5 @@ spec:
   containerName: app
 status:
   conditions:
-    - status: 'True'
+     - status: 'True'
+     - type: Completed
--- a/bundle/tests/scorecard/kuttl/day2operation/01-assert.yaml
+++ b/bundle/tests/scorecard/kuttl/day2operation/01-assert.yaml
@@ -8,4 +8,6 @@ spec:
 status:
   conditions:
      - status: 'True'
-     - type: Completed
+       type: Started
+     - status: 'True'
+       type: Completed
--- a/bundle/tests/scorecard/kuttl/day2operation/03-assert.yaml
+++ b/bundle/tests/scorecard/kuttl/day2operation/03-assert.yaml
@@ -7,4 +7,6 @@ spec:
   containerName: app
 status:
   conditions:
-    - status: 'False'
+    - reason: Error
+      status: 'False'
+      type: Started
--- a/bundle/tests/scorecard/kuttl/day2operation/03-assert.yaml
+++ b/bundle/tests/scorecard/kuttl/day2operation/03-assert.yaml
@@ -1,0 +1,10 @@
+apiVersion: rc.app.stacks/v1beta2
+kind: RuntimeOperation
+metadata:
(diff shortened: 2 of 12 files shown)


bundle/tests/scorecard/kuttl/day2operation/01-assert.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: RuntimeOperation
+metadata:
+  name: example-runtime-operation-0
+spec:
+  podName: day2-operation-rc-0
+  containerName: app
+status:
+  conditions:
+     - status: 'True'
+     - type: Completed
```
leochr, line 11: I believe both `status` and `type` should be part of the same item (remove the `-` infront of `type`). Otherwise, the test may be incorrectly marked as success when in fact it failed (i.e. `Started` type is `True`, but `Completed` is `False`)

```
status:
  conditions:
    - lastUpdateTime: '2021-11-11T17:42:52Z'
      status: 'True'
      type: Started
    - lastUpdateTime: '2021-11-11T17:42:52Z'
      status: 'False'
      type: Completed
```

bundle/tests/scorecard/kuttl/day2operation/03-assert.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: rc.app.stacks/v1beta2
+kind: RuntimeOperation
+metadata:
+  name: example-runtime-operation-1
+spec:
+  podName: day2-operation-rc-0
+  containerName: app
+status:
+  conditions:
+    - status: 'False'
```
leochr, line 10: This could be enhanced to include `type` and `reason`

```
status:
  conditions:
    - lastUpdateTime: '2021-11-17T13:22:16Z'
      message: Failed to find pod 'Specify_Pod_Name_Here' in namespace 'demo'
      reason: Error
      status: 'False'
      type: Started
```

bundle/tests/scorecard/kuttl/day2operation/04-assert.yaml
```diff
@@ -0,0 +1,22 @@
...
+kind: RuntimeOperation
+metadata:
+  name: example-runtime-operation-1
+spec:
+  podName: day2-operation-rc-0
+  containerName: app
+status:
+  conditions:
+    - status: 'True'
+    - type: Completed
```
leochr, line 22: Same comment as above (should be part of the same item)

bundle/tests/scorecard/kuttl/day2operation/05-delete-operations.yaml
```diff
@@ -0,0 +1,16 @@
+# Need to manually delete Runtime Operations as kuttl does not automatically delete these at the end of the test
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+delete:
+  - apiVersion: rc.app.stacks/v1beta2
+    kind: RuntimeOperation
+    metadata:
+      name: example-runtime-operation-0
```
leochr, line 8: This should be at the top level, similar to https://github.com/application-stacks/runtime-component-operator/blob/353c8e9189b065560d6ceb4bad8c881f5b12a694/bundle/tests/scorecard/kuttl/route-certificate/03-delete-secret.yaml#L8

Otherwise, kuttl deletes all RuntimeOperation - which is fine in this case. In that case, we would want
```
  - apiVersion: rc.app.stacks/v1beta2
    kind: RuntimeOperation
```

leochr (commented):
@yongja79 Thank you. Looks good. Added a few comments for enhancements. The tests for RuntimeOperation on a Pod that does not exist yet and then creating the pods afterwards  is great! (steps `03` and `04`)

leochr (approved):
Looks good. Thanks

Issue #263, opened 2021-11-19
Ensure that the right Pod is checked


leochr:
In kuttl test we have some assert/error statements on Pods (like [this](https://github.com/application-stacks/runtime-component-operator/blob/d1f0f946cc0e1029623f39335d07c1b827aac9a4/bundle/tests/scorecard/kuttl/annotations/01-assert.yaml#L26)), but since Pods name can vary, the name is not specified. But we can make sure that the right pod is being checked by using the labels we add:

- [x] Delivered to RCO
- [x] Delivered to OLO

leochr:
We can check for this label: `app.kubernetes.io/instance`
For example, `app.kubernetes.io/instance: runtimecomponent-sample`

PR #264, opened 2021-11-19
Ensure that the correct pods are checked in the tests

idlewis:
By adding a check for the labels on the pods
Rco fix for issue #263


leochr (approved):
@idlewis Looks good. Thanks

PR #272, opened 2021-11-23
Use build-releases to be able to detect release

hibell:

- Use correct target in makefile so that the correct tag is used for images.
- 

- [ ] User guide
- [ ] `CHANGELOG.md`

Fixes #


leochr (approved):
Thanks

PR #273, opened 2021-11-23
Minikube e2e testing

### LINKED ISSUE #235: Automation to run the e2e tests on minikube
Enable the e2e tests on minikube. We used to run the e2e tests on minikube (see the travis setup for 0.7.1 release [here](https://github.com/application-stacks/runtime-component-operator/blob/c30903ee6bb1e444dc2754f8d30274ba81eaff72/.travis.yml#L29) (e2e tests on minikube)).

- [x] Delivered minikube scripts and tests to RCO
- [x] Delivered minikube scripts and tests to OLO
- [x] Run minikube tests as part of the RCO pipeline 
- [x] Run minikube tests as part of the OLO pipeline

halim-lee:

- Support e2e testing for Minikube

- [ ] User guide
- [ ] `CHANGELOG.md`

Fixes #235


halim-lee:
Ingress addon requires drivers to be something else other than "none" - the driver is docker by default. This requires a non-root user.
For `probe` test, it runs successfully through `kubectl kuttl test ...` but it fails when it's run through operator-sdk command. I removed 02-errors.yaml and it runs successfully... This seems like kuttl's problem.

bundle/tests/scorecard/minikube-kuttl/ingress/00-assert.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: ingress
+status:
+  readyReplicas: 1
+---
+apiVersion: extensions/v1beta1
```
leochr, line 8: operator generates ingress with apiVersion `networking.k8s.io/v1`
Is this minikube specific?

scripts/installers/install-minikube.sh
```diff
@@ -10,44 +10,53 @@ export CHANGE_MINIKUBE_NONE_USER=true
...
+  echo "****** Installing Minikube v1.21.0..."
+  curl -Lo minikube https://storage.googleapis.com/minikube/releases/v1.21.0/minikube-linux-amd64 \
+  && chmod +x minikube \
+  && sudo mv minikube /usr/local/bin/
+
 
   mkdir -p $HOME/.kube $HOME/.minikube
   touch $KUBECONFIG
-  sudo minikube start --profile=minikube --vm-driver=none --kubernetes-version=v1.18.1
+  minikube start --profile=minikube --kubernetes-version=v1.20.1
```
leochr, line 44: Let's use kubectl 1.19 since OpenShift 4.6 (Long supported version) is at that level.

leochr (commented):
@halim-lee Thanks for the PR. Looks good. Since the majority of the tests are the same between ocp and minikube, it'll be good to add script to copy the existing tests, remove the ones that are not needed for minikube (i.e. route) and then add the tests for minikube (ingress).

bundle/tests/scorecard/minikube-kuttl/ingress/00-assert.yaml
halim-lee: Hmm.. `apiVersion: extensions/v1beta1` is what gets created in the Minikube cluster when expose is set to True. Minikube lists both `networking.k8s.io/v1` and `extensions/v1beta1` as supported.

PR #291, opened 2021-11-30
Add kustomize files for 0.8.0 release

idlewis:
Also update the kustomize CRDs to pickup the most
recent changes

As per the request in issue #177


deploy/releases/0.8.0/kustomize/base/runtime-component-operator.yaml
```diff
@@ -0,0 +1,237 @@
...
+        - --enable-leader-election
+        command:
+        - /manager
+        env:
+        - name: OPERATOR_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        - name: WATCH_NAMESPACE
+          value: default
```
leochr, line 225: Value for WATCH_NAMESPACE can also be specified like OPERATOR_NAMESPACE. In that case the example for watching own namespace can be simplified I think.

fyi to avoid merge conflicts, Melissa has the change to update kustomize/daily/runtime-component-operator.yaml in her PR

leochr (commented):
@idlewis Thanks! Added a minor comment for the watch-own-namespace scenario. Have some questions about the watch-another-namespace scenario, let's chat tomorrow.

idlewis:
@leochr I've addressed the comments about the watch-own-namespace scenario

deploy/kustomize/daily/examples/watch-another-namespace/rco-deployment.yaml
```diff
@@ -11,3 +11,4 @@ spec:
           env:
             - name: WATCH_NAMESPACE
               value: rco-watched-ns2
+              valueFrom:
```
leochr, line 14: Since this example builds upon overlay, which already removes the `valueFrom` field, it doesn't need to be specified here again.

leochr (commented):
@idlewis Thanks for the update. One minor comment about the examples.

leochr (approved):
Looks good. Thanks

PR #293, opened 2021-12-01
Allow insecureEdgeTerminationPolicy set to redirect on passthrough

yongja79:
…n Passthrough

- If TLS termination is set to Passthrough we need to also set InsecureEdgeTerminationPolicy as Redirect is a possible option

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -126,7 +126,11 @@ func CustomizeRoute(route *routev1.Route, ba common.BaseComponent, key string, c
 				route.Spec.TLS.CACertificate = ""
 				route.Spec.TLS.Key = ""
 				route.Spec.TLS.DestinationCACertificate = ""
-				route.Spec.TLS.InsecureEdgeTerminationPolicy = ""
+				if *rt.GetInsecureEdgeTerminationPolicy() == routev1.InsecureEdgeTerminationPolicyRedirect {
+					route.Spec.TLS.InsecureEdgeTerminationPolicy = *rt.GetInsecureEdgeTerminationPolicy()
+				} else {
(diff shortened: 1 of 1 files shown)


yongja79:
Required for OLO issue https://github.com/OpenLiberty/open-liberty-operator/issues/297

utils/utils.go
```diff
@@ -126,7 +126,11 @@ func CustomizeRoute(route *routev1.Route, ba common.BaseComponent, key string, c
 				route.Spec.TLS.CACertificate = ""
 				route.Spec.TLS.Key = ""
 				route.Spec.TLS.DestinationCACertificate = ""
-				route.Spec.TLS.InsecureEdgeTerminationPolicy = ""
+				if *rt.GetInsecureEdgeTerminationPolicy() == routev1.InsecureEdgeTerminationPolicyRedirect {
```
leochr, line 129: @arturdzm any particular reason why the InsecureEdgeTerminationPolicy was not being set for passthrough termination? It seems like Redirect and None are allowed values.

leochr: @arturdzm please review the above question when you get a moment. Thanks.

leochr: @yongja79 Need a nil check since InsecureEdgeTerminationPolicy could have not been specified in CR. Also, `Allow` is not a valid value for `passthrough`. The operator should throw a validation error (Reconcile should fail and output a relevant message).

leochr: OpenShift UI does provide Redirect and None as options for passthrough

passthrough

leochr (approved):

PR #302, opened 2021-12-17
Add kustomization patch to add description for service binding secret

idlewis:
This means that the open shift UI for a RuntimeComponent will render and link to the service binding secret

This address issue #289


leochr (approved):
@idlewis Thank you. Looks good! Please run `make bundle` and check in the resulting changes as well.

PR #309, opened 2021-12-23
Auto update containerImage and createdAt when building the bundle

### LINKED ISSUE #292: Update CSV annotation values automatically
The base for the ClusterServiceVersion ([here](https://github.com/application-stacks/runtime-component-operator/blob/44c89e64bf9f8b73e4e95ca1197e14e1f3833145/config/manifests/bases/runtime-component.clusterserviceversion.yaml#L8)) specifies these annotations that are used by OpenShift UI to render some information about the operator:

These values are managed manually. It'll be good to automate that as part of `make bundle`. So that the release builds can also update the values on the fly (pipeline scripts call `make bundle` which update the actual image defined for Deployment)

    createdAt: "2021-11-25T14:00:00Z"
    containerImage: applicationstacks/operator:daily

- [x] Delivered to RCO
- [x] Delivered to OLO

yongja79:

- When make bundle is called updates the csv's containerImage and createdAt values automatically to match those in the Makefile

Fixes #292


leochr (approved):
Thanks for the update

PR #312, opened 2022-01-10
Update makefile sed command for linux

yongja79:

- Fix the sed command for the auto update of the containerImage to work on linux for the build


leochr (approved):
@yongja79 Looks good. Thank you

PR #322, opened 2022-01-12
Replace wildcards with specific permission verbs

### LINKED ISSUE #255: Specify verbs explicitly for permissions instead of wildcards
RBAC permissions specify wildcards for verbs (an example is [here](https://github.com/application-stacks/runtime-component-operator/blob/8f629213476c981223cca9f16b7a5dd6de5c432f/bundle/manifests/runtime-component.clusterserviceversion.yaml#L473)). We should explicitly specify the list of verbs instead in order to remove all wildcards.

Above values in CSV are generated automatically based on the `kubebuilder:rbac` annotations specified in the controllers (an example is [here](https://github.com/application-stacks/runtime-component-operator/blob/807dc286a8a4909028553fa13c808040add9f89f/controllers/runtimecomponent_controller.go#L65)). So those should be modified accordingly.

   - [x] Delivered to RCO
   - [ ] Delivered to OLO

halim-lee:
- Specifying list of verbs explicitly instead of wildcards

- [ ] User guide
- [ ] `CHANGELOG.md`

### DIFF
--- a/bundle/manifests/runtime-component.clusterserviceversion.yaml
+++ b/bundle/manifests/runtime-component.clusterserviceversion.yaml
@@ -38,7 +38,7 @@ metadata:
     categories: Application Runtime
     certified: "true"
     containerImage: applicationstacks/operator:daily
-    createdAt: "2021-11-25T14:00:00Z"
+    createdAt: 22-01-24T14:41:26Z
     description: Deploys any runtime component with dynamic and auto-tuning configuration
     operators.operatorframework.io/builder: operator-sdk-v1.6.1+git
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
@@ -487,7 +487,13 @@ spec:
           - deployments
           - statefulsets
           verbs:
-          - '*'
+          - create
+          - delete
+          - get
+          - list
+          - patch
+          - update
+          - watch
         - apiGroups:
           - apps
           resources:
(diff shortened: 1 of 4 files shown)


bundle/manifests/runtime-component.clusterserviceversion.yaml
```diff
@@ -487,7 +487,13 @@ spec:
           - deployments
           - statefulsets
           verbs:
-          - '*'
+          - create
```
leochr, line 490: There are 2 more wildcards in RuntimeOperation [controller](https://github.com/application-stacks/runtime-component-operator/blob/main/controllers/runtimeoperation_controller.go).

config/manifests/patches/csvAnnotations.yaml
```diff
@@ -4,5 +4,5 @@ metadata:
   name: runtime-component.v0.0.0
   namespace: placeholder
   annotations:
-    containerImage: IMAGE
-    createdAt: CREATEDAT
\ No newline at end of file
+    containerImage: applicationstacks/operator:daily
```
leochr, line 7: Please rebase and run the `make bundle` command. With the latest changes, this template file shouldn't be modified.

leochr (commented):
@halim-lee Thanks for the PR. Added a couple of comments.

leochr:
Fixes #255 (adding as a comment, otherwise merging the PR would close the issue, but need the issue to remain open until changes are delivered to OLO as well)

Issue #305, opened 2021-12-18
Add tests for initContainers and sidecarContainers


leochr:
Init containers and sidecar containers can be specified for the Deployment and StatefulSet using these fields in CR:
- spec.initContainers
- spec.sidecarContainers

Add tests to ensure these are added to Deployment/StatefulSet as expected and removed when the relevant config is removed from CR.

- [x] Delivered to RCO
- [x] Delivered to OLO
- [x]  Delivered to WLO

PR #323, opened 2022-01-13
Add tests for sidecar containers

idlewis:
Sidecar container tests as per issue #305

### DIFF
--- a/bundle/tests/scorecard/kuttl/sidecar-deployment/02-assert.yaml
+++ b/bundle/tests/scorecard/kuttl/sidecar-deployment/02-assert.yaml
@@ -1,0 +1,5 @@
+apiVersion: v1
+kind: Pod
+spec:
+  containers:
+  - name: app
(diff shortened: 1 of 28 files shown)


bundle/tests/scorecard/kuttl/sidecar-deployment/02-assert.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: Pod
+spec:
+  containers:
+  - name: app
```
leochr, line 5: It'll be good to add the label `app.kubernetes.io/instance` to ensure that it's the right Pod and not a Pod leftbehind from previous test scenarios (i.e. init-container)

bundle/tests/scorecard/kuttl/sidecar-stateful/02-assert.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: Pod
+spec:
+  containers:
+  - name: app
```
leochr, line 5: Same comment about Pod label.

leochr (commented):
@idlewis Looks great! Thanks. Added a minor comment about using label for certain Pods to ensure it was the right Pod.

idlewis:
@leochr updated as requested

leochr (approved):
Thanks

Issue #313, opened 2022-01-10
Ensure volume mount and env are specified for application container


leochr:
[service-certificate test](https://github.com/application-stacks/runtime-component-operator/blob/main/bundle/tests/scorecard/kuttl/service-certificate/01-assert.yaml) checks that the volume _svc-certificate_ is specified in _Deployment_. In addition, it should also check for volumeMount (with the same name) and env named `TLS_DIR`. Relevant code that adds these is [here](https://github.com/application-stacks/runtime-component-operator/blob/f8a61ea2ebeb27c5ad20cacd9285b18f03e5a645/utils/utils.go#L420).

- [ ] Delivered to RCO
- [ ] Delivered to OLO

PR #324, opened 2022-01-14
Improve the service-certificate test

idlewis:
Add checks for the TLS_DIR env var and the
certificate volume mount

This is to deal with issue #313


bundle/tests/scorecard/kuttl/service-certificate/01-assert.yaml
```diff
@@ -9,6 +9,13 @@ status:
 spec:
   template:
     spec:
+      containers:
+      - env:
+        - name: TLS_DIR
+          value: /etc/x509/certs
+        volumeMounts:
+        - name: svc-certificate
+          mountPath: /etc/x509/certs
```
leochr, line 18: It'll be good to ensure that read-only is set to true as well.

leochr (commented):
@idlewis Thanks for the PR. Looks good. Just a minor comment about the read-only field.

idlewis:
@leochr updated as requested

PR #337, opened 2022-02-23
Update RBAC permission for rco-manager-role

### LINKED ISSUE #330: Review permissions and adjust accordingly
Review the permissions (RBAC) set for the resources and scope down the verbs where possible. Needs to be done carefully, otherwise, the Operator wouldn't be able to do the necessary operation(s).

yongja79:

- Removes unnecessary permissions given to the rco-manager-role

Fixes #330


controllers/runtimecomponent_controller.go
```diff
@@ -61,16 +61,16 @@ type RuntimeComponentReconciler struct {
...
+// +kubebuilder:rbac:groups=apps,resources=deployments;statefulsets,verbs=list;watch;create;update;delete,namespace=runtime-component-operator
 // +kubebuilder:rbac:groups=apps,resources=deployments/finalizers;statefulsets,verbs=update,namespace=runtime-component-operator
-// +kubebuilder:rbac:groups=core,resources=services;secrets;serviceaccounts;configmaps,verbs=get;list;watch;create;update;patch;delete,namespace=runtime-component-operator
-// +kubebuilder:rbac:groups=autoscaling,resources=horizontalpodautoscalers,verbs=get;list;watch;create;update;patch;delete,namespace=runtime-component-operator
-// +kubebuilder:rbac:groups=networking.k8s.io,resources=ingresses,verbs=get;list;watch;create;update;patch;delete,namespace=runtime-component-operator
-// +kubebuilder:rbac:groups=route.openshift.io,resources=routes;routes/custom-host,verbs=get;list;watch;create;update;patch;delete,namespace=runtime-component-operator
+// +kubebuilder:rbac:groups=core,resources=services;secrets;serviceaccounts,verbs=list;watch;create;update;delete,namespace=runtime-component-operator
```
leochr, line 67: @yongja79 Operator does interact with at least one ConfigMap that has the optional configurations. Relevant controller code is [here](https://github.com/application-stacks/runtime-component-operator/blob/66fe693b8afc7d38b92b68f4a72c14bd7711152a/controllers/runtimecomponent_controller.go#L99)

It only logs the error and proceeds. @arturdzm is that behaviour intentional? Shouldn't the ConfigMap always exist?

yongja79: @leochr I've added back the configMap permissions to the rco-manager role and have also updated the rco-leader-election role

config/rbac/leader_election_role.yaml
```diff
@@ -9,21 +9,19 @@ rules:
   resources:
   - configmaps
   verbs:
+  - create
```
leochr, line 12: @yongja79 Similar to WLO, changes to leader election should be reversed. Thanks.

leochr (approved):

PR #343, opened 2022-03-07
Dont create deployment without a pull secret

idlewis:
The cluster automatically creates pull secrets for the service account.
This can take a while.
Sometimes, the controller creates the deployment before the pull secret
is created. This causes the pods to go into an image pull backoff loop.
Check that the pull secret exists before the deployment is created, to
prevent this

This will help us to fix issue OpenLiberty/open-liberty-operator#312

### DIFF
--- a/controllers/runtimecomponent_controller.go
+++ b/controllers/runtimecomponent_controller.go
@@ -221,6 +221,13 @@ func (r *RuntimeComponentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		}
 	}
 
+	// Check if the ServiceAccount has a valid pull secret before creating the deployment/statefulset
+	// or setting up knative. Otherwise the pods can go into an ImagePullBackOff loop
+	saErr := appstacksutils.ServiceAccountPullSecretExists(instance, r.GetClient())
+	if saErr != nil {
+		return r.ManageError(saErr, common.StatusConditionTypeReconciled, instance)
+	}
+
 	isKnativeSupported, err := r.IsGroupVersionSupported(servingv1.SchemeGroupVersion.String(), "Service")
 	if err != nil {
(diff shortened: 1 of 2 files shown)


controllers/runtimecomponent_controller.go
```diff
@@ -221,6 +221,17 @@ func (r *RuntimeComponentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		}
 	}
 
+	// Check if the ServiceAccount has a valid pull secret before creating the deployment/statefulset
+	// or setting up knative. Otherwise the pods can go into an ImagePullBackOff loop
+	saName := instance.Name
+	if instance.Spec.ServiceAccountName != nil && *instance.Spec.ServiceAccountName != "" {
+		saName = *instance.Spec.ServiceAccountName
+	}
+	saErr := appstacksutils.PullSecretExists(saName, instance.Namespace, r.GetClient())
```
leochr, line 230: To minimize the amount of changes needed for each controller, it'll be good if the above logic to determine the service account name is also moved into the util method. The variable `ba` (of type common.BaseComponent) can be passed instead of _saName_ and _instance.Namespace_. They can be retrieved from `ba` along with CR fields (i.e. ba.GetServiceAccountName()). Util method [CustomizeRoute](https://github.com/application-stacks/runtime-component-operator/blob/db4d0420b902d713a0b3e0e2f6197c9addbb83a4/utils/utils.go#L96) is a good reference. It's fine to pass in client along with ba.

Suggested name for the method: ServiceAccountPullSecretExists

leochr (commented):
@idlewis Looks good. Thank you. Added a comment to move more code into util.

idlewis:
@leochr review comments addressed, pls could you take another look?

leochr (approved):
@idlewis Looks good. Thank you.

PR #357, opened 2022-03-30
Enable TLS by default

arturdzm:

-  Adds spec.manageTLS property (default true)
-  when manageTLS is true uses cert-manager or OCP to generate pod certificates
-  Adds relevant certificate info into Prometheus Service Monitor if enabled


utils/utils.go
```diff
@@ -731,6 +740,19 @@ func CustomizeServiceMonitor(sm *prometheusv1.ServiceMonitor, ba common.BaseComp
 		sm.Spec.Endpoints[0].MetricRelabelConfigs = endpoints[0].MetricRelabelConfigs
 		sm.Spec.Endpoints[0].HonorTimestamps = endpoints[0].HonorTimestamps
 		sm.Spec.Endpoints[0].HonorLabels = endpoints[0].HonorLabels
+	}
+	if ba.GetManageTLS() == nil || *ba.GetManageTLS() {
```
leochr, line 744: ServiceMonitor is always created from scratch, so no clean-up is needed. Just wanted to confirm if that's the case?

leochr (commented):
@arturdzm Thanks for the PR. Looks good. Added some minor comments and questions.

utils/reconciler.go
```diff
@@ -357,3 +357,136 @@ func (r *ReconcilerBase) GetRouteTLSValues(ba common.BaseComponent) (key string,
...
+			cleanup()
+			return false, nil
+		}
+		if _, ok := ba.GetService().GetAnnotations()["service.alpha.openshift.io/serving-cert-secret-name"]; ok {
+			cleanup()
+			return false, nil
+		}
+	}
+	if ok, err := r.IsGroupVersionSupported(certmanagerv1.SchemeGroupVersion.String(), "Certificate"); err != nil {
+		return false, err
```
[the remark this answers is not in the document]
arturdzm: This is one should be fine to leave as is. As this error means failed to check (permission, connection, etc) not that it is not installed

leochr (commented):
@arturdzm Thanks for the updates. Just a couple more comments.

leochr:
@arturdzm Some of the probe tests recently delivered are failing in the PR build. At this particular step, which uses a new sample Liberty app, which is what we'll be switching to as the default CR in the CSV.
https://github.com/application-stacks/runtime-component-operator/blob/main/bundle/tests/scorecard/kuttl/probe/03-runtime-probe-defaults.yaml
Could you please investigate why the probe fails with manageTLS? I tired it on the canopus cluster. I'll slack you the link to it. The pod is restarting continiously.

utils/utils.go
```diff
@@ -1183,3 +1208,68 @@ func getSecurityContext(ba common.BaseComponent) *corev1.SecurityContext {
...
 	return secContext
 }
+
+func AddOCPCertAnnotation(ba common.BaseComponent, svc *corev1.Service) {
+	bao := ba.(metav1.Object)
+
+	if ba.GetCreateKnativeService() != nil && *ba.GetCreateKnativeService() {
+		if val, _ := svc.Annotations["service.beta.openshift.io/serving-cert-secret-name"]; val == bao.GetName()+"-svc-tls-ocp" {
+			delete(ba.GetStatus().GetReferences(), common.StatusReferenceCertSecretName)
+			delete(ba.GetStatus().GetReferences(), "service.beta.openshift.io/serving-cert-signed-by")
```
leochr, line 1218: Should be deleted from svc.Annotations

`service.beta.openshift.io/serving-cert-secret-name` should also be deleted

utils/utils.go
```diff
@@ -1183,3 +1208,68 @@ func getSecurityContext(ba common.BaseComponent) *corev1.SecurityContext {
...
+			delete(ba.GetStatus().GetReferences(), "service.alpha.openshift.io/serving-cert-signed-by")
+
+		}
+		return
+	}
+
+	if ba.GetManageTLS() != nil && !*ba.GetManageTLS() || ba.GetService() != nil && ba.GetService().GetCertificateSecretRef() != nil {
+		if val, _ := svc.Annotations["service.beta.openshift.io/serving-cert-secret-name"]; val == bao.GetName()+"-svc-tls-ocp" {
+			delete(ba.GetStatus().GetReferences(), common.StatusReferenceCertSecretName)
+			delete(ba.GetStatus().GetReferences(), "service.beta.openshift.io/serving-cert-signed-by")
```
leochr, line 1228: Same here

common/config.go
```diff
@@ -8,18 +8,15 @@ import (
...
-
-	// OpConfigSvcBindingGVKs a comma-seperated list of GroupVersionKind values in "<Kind>.<version>.<group>" format.
-	// e.g. "CronTab.v1.stable.example.com"
-	//OpConfigSvcBindingGVKs = "serviceBinding.groupVersionKinds"
 
 	// OpConfigDefaultHostname a DNS name to be used for hostname generation.
 	OpConfigDefaultHostname = "defaultHostname"
+
+	// OpConfigCMCADuration default duration for cert-manager issued CA
+	OpConfigCMCADuration = "cmCADuration"
```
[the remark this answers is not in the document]
leochr: Still seeing the old value for these

leochr (commented):
@arturdzm some minor updates needed, thanks

leochr (approved):
@arturdzm Looks good! Thanks

PR #361, opened 2022-04-02
fix: edits resource limit for runtime component manager operator container from 256Mi to 512Mi (#359)

### LINKED ISSUE #359: Resource limits for operator to low in ClusterServiceVersion
https://github.com/application-stacks/runtime-component-operator/blob/044b87e47baa346601385e7c69059065ecb77bfe/bundle/manifests/runtime-component.clusterserviceversion.yaml#L440

In our case on openshift this is not enough and our RCO Operator keeps crashing.. So we need to adjust this somehow manually.. also not sure how to do that, thats its persisted when operator upgrades..

We are getting OOMKill, so explicitly memory limits seems to low.

Elyytscha:

This PR changes resource memory limit for rco controller manager from 256Mi to 512Mi
Because on Openshift, recent rco controller runs into OOMKill (0.8.0, 0.8.1 tested)

No

Fixes #359

### DIFF
--- a/bundle/manifests/runtime-component.clusterserviceversion.yaml
+++ b/bundle/manifests/runtime-component.clusterserviceversion.yaml
@@ -437,7 +437,7 @@ spec:
                 resources:
                   limits:
                     cpu: 200m
-                    memory: 256Mi
+                    memory: 512Mi
                   requests:
                     cpu: 100m
                     memory: 64Mi


leochr:
@Elyytscha Thanks for the PR. For some reason(s), we didn't encounter OOM issue during our tests, even with some minimal load testing. So wondering how many instances of RuntimeComponents and/or RuntimeOperations were being managed by the operator when the issue occurs?

leochr:
@Elyytscha Thanks for the PR. For some reason(s), we didn't encounter OOM issue during our tests, even with some minimal load testing. So wondering how many instances of RuntimeComponents and/or RuntimeOperations were being managed by the operator when the issue occurs?

Elyytscha:
When we upgraded the operator to 0.8.1, there where 3 RuntimeComponent CR's when the issue appeared.

Elyytscha:
@leochr are there some updates? would be nice if this could be included in 0.8.2 because we don't know how we can upgrade our runtime component operator without breaking it for a few minutes, where we have to manually adjust this limits. as you can see in the screenshot our rco manager instance is using more then 400 MB RAM, so 256MB is definitely to low.

PR #368, opened 2022-04-06
Set default affinity rules if no user affinity set

### LINKED ISSUE #354: Add affinity rules by default to schedule pods on different nodes across different zones
By default, we should add perferred podAntiAffinity rules to spread the pods across different nodes and zones. This would help with high availability out of the box. Users can override with the `.spec.affinity.*` fields.

yongja79:

- Sets default affinity rules if the user has not set any
- Any user supplied rules will overwrite the defaults
- Add to the affinity tests to cover this change

- [ ] User guide
- [ ] `CHANGELOG.md`

Fixes #354


bundle/tests/scorecard/kuttl/affinity/08-default-affinity.yaml
```diff
@@ -0,0 +1,14 @@
+# Test to see that the default affinity is is overwritten if supplied with a user defined affinity
```
leochr, line 1: An error check could be added for default podAntiAffinity

leochr, line 1: An error check could be added for default podAntiAffinity

leochr (commented):
@yongja79 Thanks for the PR. Added a couple of minor comments.

leochr (approved):
@yongja79 Looks good. Please update WLO as well. Thanks.

Issue #350, opened 2022-03-16
Address pull secret synchronization issues


leochr:
Users can specify image pull secrets (IPS) using the `.spec.pullSecret` field. The pull secret is added to the ServiceAccount (SA) that the operator creates and eventually set on the Pods.

- [x] Create the CR. Once Deployment is created, specify `.spec.pullSecret` in CR. The pull secret value is not added to Pod. If Pod is deleted then the value is added.
- [x] Create a new CR with `.spec.pullSecret` set to a Secret that doesn't exist. Reconciled status should report an error. ~Update CR to specify a valid Secret.~ **Remove the pull secret.** The error doesn't go away and Deployment is never created. This is because the IPS value is never removed from SA and to fix the ImagePullBackOff issue, we check the first IPS value in SA to make sure Secret exists.

leochr:
@idlewis I probably misremembered the details. Anyways, here is the scenario that is not working as expected:
Create CR with an invalid pullSecret value. Reconciled reports an error. Remove the pullSecret. The error doesn't go away. That's because the image pull secret value is not removed from the service account. We don't know which pull secret to remove. Tracking the pullSecret in status would allow us to remove it.

leochr:
WLO: https://github.com/WASdev/websphere-liberty-operator/pull/139

PR #369, opened 2022-04-07
Ensure pods are updated if the service account is

idlewis:
This deals with #350

After this change, the pods will be deleted/recreated automatically if a pullSecret is added to the CR

Also, the pull secret will be deleted from the service account if it is deleted from the CR.


idlewis:
@halim-lee This is the PR for issue #350

utils/utils.go
```diff
@@ -591,6 +591,13 @@ func CustomizePodSpec(pts *corev1.PodTemplateSpec, ba common.BaseComponent) {
 		})
 	}
 
+	// This ensures that the pods are updated if the service account is updated
```
leochr, line 594: The pods should be updated for user-provided service account as well. Setting the reference value at the end of `ServiceAccountPullSecretExists` would ensure that the right service account is checked for existence, it has image pull secret (if it exists) and finally set the resource version of the corresponding service account.

utils/utils.go
```diff
@@ -673,6 +686,15 @@ func CustomizeServiceAccount(sa *corev1.ServiceAccount, ba common.BaseComponent)
 	}
 }
 
+func removePullSecret(sa *corev1.ServiceAccount, pullSecretName string) {
+	for i, obj := range sa.ImagePullSecrets {
+		if obj.Name == pullSecretName {
+			sa.ImagePullSecrets = append(sa.ImagePullSecrets[:i], sa.ImagePullSecrets[i+1:]...)
```
leochr, line 692: I am not entirely sure if a concurrent modification exception could happen in Go for this scenario (iterating through pull secrets and setting a value for the same). If we are not sure, getting the index of the pull secret we are looking for and setting this outside of for-loop is another option.

utils/utils.go
```diff
@@ -662,7 +669,13 @@ func CustomizeServiceAccount(sa *corev1.ServiceAccount, ba common.BaseComponent)
...
 	sa.Annotations = MergeMaps(sa.Annotations, ba.GetAnnotations())
 
+	psr := ba.GetStatus().GetReferences()[common.StatusReferencePullSecretName]
+	if psr != "" && (ba.GetPullSecret() == nil || *ba.GetPullSecret() != psr) {
+		removePullSecret(sa, psr)
+	}
+
 	if ba.GetPullSecret() != nil {
+		ba.GetStatus().SetReference(common.StatusReferencePullSecretName, *ba.GetPullSecret())
 		if len(sa.ImagePullSecrets) == 0 {
```
leochr, line 679: Adding the pull secret by iterating through the secrets would be better than adding it based on index. WSA for [reference](https://github.ibm.com/websphere/automation-operator/blob/ab96861893c32f30f9918ee8d939db7b7c6fcaa6/controllers/websphereautomation_controller.go#L543)

leochr (commented):
@idlewis thanks for the PR and @halim-lee for the updates. Added some comments.

utils/utils.go
```diff
@@ -662,7 +669,13 @@ func CustomizeServiceAccount(sa *corev1.ServiceAccount, ba common.BaseComponent)
 	sa.Labels = ba.GetLabels()
 	sa.Annotations = MergeMaps(sa.Annotations, ba.GetAnnotations())
 
+	psr := ba.GetStatus().GetReferences()[common.StatusReferencePullSecretName]
+	if psr != "" && (ba.GetPullSecret() == nil || *ba.GetPullSecret() != psr) {
+		removePullSecret(sa, psr)
```
leochr, line 674: @idlewis `common.StatusReferencePullSecretName` should also be unset once the pull secret is removed, so this if-condition is not continuously executed.

idlewis:
closing in favour of PR #387 as I've tweaked the commits a fair bit. All the review comments from this PR are addressed in 387

PR #328, opened 2022-01-19
Add tests for service types

yongja79:
Signed-off-by: Jason Yong <jason_yong@uk.ibm.com>

**What this PR does / why we need it?**:

- Added service tests for the ClusterIp, LoadBalancer and NodePort types as well as other fields not currently tested

PR #373, opened 2022-04-12
If targetport not set default it to the port value within the CR

### LINKED ISSUE #346: TargetPort under spec.service.ports is set to 0 when not defined
### What did you do?

A clear and concise description of the steps you took (or insert a code snippet).

Defined `spec.service.ports.port` and left `spec.service.ports.targetPort` blank. After reconciliation, targetPort appears to be 0 in RuntimeComponent CR. 
```
spec:
  applicationImage: 'openliberty/samples:springPetClinic'
  applicationName: test-app
  pullPolicy: IfNotPresent
  replicas: 4
  resources: {}
  service:
    port: 9080
    ports:
      - port: 8080
        protocol: TCP
        targetPort: 0
    type: ClusterIP
```

Service actually does assign the correct value to targetPort (same value as port).
```
spec:
  ports:
    - name: 9080-tcp
      protocol: TCP
      port: 9080
      targetPort: 9080
    - name: 8080-tcp
      protocol: TCP
      port: 8080
      targetPort: 8080
```

Looks like targetPort in RuntimeComponent CR does not get updated properly.

A clear and concise description of what you expected to happen (or insert a code snippet). If possible, add `RuntimeComponent` CR.

```
spec:
  applicationImage: 'openliberty/samples:springPetClinic'
  applicationName: test-app
  pullPolicy: IfNotPresent
  replicas: 4
  resources: {}
  service:
    port: 9080
    ports:
      - port: 8080
        protocol: TCP
        targetPort: 8080
    type: ClusterIP
```

yongja79:

- If the Targetports are not set within a ServicePort then the CR sets and displays them as 0 even though the service correctly sets them to the Port value. This PR ensures that the CR value for Targetports match that of the Service created.

Fixes #346

### DIFF
--- a/controllers/runtimecomponent_controller.go
+++ b/controllers/runtimecomponent_controller.go
@@ -51,6 +51,7 @@ import (
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	servingv1 "knative.dev/serving/pkg/apis/serving/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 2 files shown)


yongja79:
@leochr Rather than adding a separate test for this I will update the test case in [PR328 ](https://github.com/application-stacks/runtime-component-operator/pull/328) as they are all about services

controllers/runtimecomponent_controller.go
```diff
@@ -146,7 +147,13 @@ func (r *RuntimeComponentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		// it is not possible to override converted annotations.
 		instance.Annotations = appstacksutils.MergeMaps(instance.Annotations, appstacksutils.GetOpenShiftAnnotations(instance))
 	}
-
+	// If TargetPorts on Serviceports are not set, default them to the Port value in the CR instance
```
leochr, line 150: It'll be good to do this at the end of the Initialize() method called above. If the actual logic to go through the ports is added to a common util method, other operators can also invoke it from their Initialize methods.

leochr (commented):
@yongja79 Thanks for the PR. The changes to set target port value is good. Added a comment to restructure the code.

controllers/runtimecomponent_controller.go
yongja79: @leochr  I've move the logic into the initialize() method but couldn't move it into the common utils as it would create a circular import

leochr (approved):
@yongja79 Looks good. Thanks.

PR #375, opened 2022-04-18
Allow ingress traffic from OpenShift router and monitoring pods

hibell:

- OpenShift: New default behavior for exposed apps is to only allow traffic from the OpenShift router
- OpenShift: Allow traffic from OpenShift monitoring pods
- Added `namespaceLabels` to allow users to override the namespace labels for ingress traffic

- [ ] User guide
- [ ] `CHANGELOG.md`

Fixes #


utils/utils.go
```diff
@@ -325,28 +325,55 @@ func CustomizeNetworkPolicy(networkPolicy *networkingv1.NetworkPolicy, ba common
 		networkPolicy.Spec.Ingress[0].From = append(networkPolicy.Spec.Ingress[0].From, networkingv1.NetworkPolicyPeer{})
```
leochr, line 325: By creating a new Ingress and setting it to `networkPolicy.Spec.Ingress`, the `len` checks above and below can be removed. The Ingress rule would always be created from scratch based on the CR values.

utils/utils.go
```diff
@@ -325,28 +325,55 @@ func CustomizeNetworkPolicy(networkPolicy *networkingv1.NetworkPolicy, ba common
...
-	if exposed := ba.GetExpose(); exposed != nil && *exposed {
-		customizeExposedNetworkPolicyPeer(&networkPolicy.Spec.Ingress[0])
+	networkPolicyConfig := ba.GetNetworkPolicy()
+	if isNetworkPolicyConfigSet(networkPolicyConfig) {
+		customizeNetworkPolicyPeer(&networkPolicy.Spec.Ingress[0].From[0], networkPolicyConfig)
+	} else if isExposed := ba.GetExpose(); isExposed != nil && *isExposed {
+		customizeExposedNetworkPolicyPeer(&networkPolicy.Spec.Ingress[0].From[0], isOpenShift)
 	} else {
-		customizeNetworkPolicyPeer(networkPolicy.Namespace, &networkPolicy.Spec.Ingress[0], ba)
+		customizeInternalNetworkPolicyPeer(&networkPolicy.Spec.Ingress[0].From[0], networkPolicy.Namespace, ba)
```
leochr, line 334: The logic in customizeNetworkPolicyPeer and customizeInternalNetworkPolicyPeer probably should be combined. - Use `"app.kubernetes.io/part-of"` as pod selector label by default. If fromLabels is specified in CR, use it instead. 
- Restrict to CR namespace by default. If namespaceLabels is specified, specify it instead.
- both can be set to `{}` to allow all pods/namespaces

leochr (commented):
@hibell Thank you for the PR! Added some comments. I've booked a meeting tomorrow (Tuesday) at 9:30 am EDT to go through the comments and discuss further.

bundle/tests/scorecard/kuttl/network-policy/07-assert.yaml
```diff
@@ -0,0 +1,27 @@
...
+timeout: 60
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: network-policy-rc
+spec:
+  ingress:
+    - from:
+        - namespaceSelector:
```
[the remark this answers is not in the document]
hibell: Good catch. It looks like using `{}` didn't result in the CR being patched as intended.

bundle/tests/scorecard/kuttl/network-policy/05-errors.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: networking.k8s.io/v1
```
leochr, line 1: @hibell should this be removed?

leochr (approved):
@hibell Looks good! Thanks

PR #377, opened 2022-04-18
Check for secrets only if service account pull secrets exist

halim-lee:
- The operator checks if secret resource exists even if service account does not have pull secret added. Check only if pull secret is added to service account.

- [ ] User guide
- [ ] `CHANGELOG.md`

Related: https://github.com/OpenLiberty/open-liberty-operator/issues/340


leochr (approved):
@halim-lee Looks good. Thanks

PR #383, opened 2022-05-03
Update env var for operator namespace

smcclem:

As the result of a bug, the operator is unable to create the "runtime-component-operator" ConfigMap in any other namespace other than the first namespace listed in the watch namepace(s) list. If all namespaces are being watched, then no ConfigMap would be created.    Update the code to read the namespace the operator is running and use that namespace to check for and create the runtime-component-operator ConfigMap if necessary. 

- 
- 

- [ ] User guide
- [ ] `CHANGELOG.md`

Fixes #


leochr (approved):
@smcclem Looks good. Thank you

PR #387, opened 2022-05-11
 Ensure pods are updated if the service account is

idlewis:
This deals with https://github.com/application-stacks/runtime-component-operator/issues/350

After this change, the pods will be deleted/recreated automatically if a pullSecret is added to the CR

Also, the pull secret will be deleted from the service account if it is deleted from the CR.


idlewis:
I'm closing pr #369 in favour of this as I've tweaked the commits a fair bit.
All the review comments from that PR have been addressed

leochr (approved):
Looks good. Thanks @idlewis and @hibell

Issue #389, opened 2022-05-13
Pull secret specified in CR is not validated


leochr:
Check and validate that the pull secret specified in CR at `.spec.pullSecret` exists. If not, throw a reconcile error.

PR #391, opened 2022-05-19
Verify pull secret exists

idlewis:
If a pull secret is specified in the CR, verify that it exists
before adding it to the service account

Addresses application-stacks/runtime-component-operator#389

### DIFF
--- a/utils/reconciler_test.go
+++ b/utils/reconciler_test.go
@@ -76,7 +76,7 @@ func TestCreateOrUpdate(t *testing.T) {
 	r := NewReconcilerBase(rcl, cl, s, &rest.Config{}, record.NewFakeRecorder(10))
 
 	err := r.CreateOrUpdate(serviceAccount, runtimecomponent, func() error {
-		CustomizeServiceAccount(serviceAccount, runtimecomponent)
+		CustomizeServiceAccount(serviceAccount, runtimecomponent, r.GetClient())
 		return nil
 	})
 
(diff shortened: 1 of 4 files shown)


utils/reconciler_test.go
```diff
@@ -76,7 +76,7 @@ func TestCreateOrUpdate(t *testing.T) {
 	r := NewReconcilerBase(rcl, cl, s, &rest.Config{}, record.NewFakeRecorder(10))
 
 	err := r.CreateOrUpdate(serviceAccount, runtimecomponent, func() error {
-		CustomizeServiceAccount(serviceAccount, runtimecomponent)
+		CustomizeServiceAccount(serviceAccount, runtimecomponent, r.GetClient())
```
leochr, line 79: should it be return CustomizeServiceAccount(....) ? or it doesn't matter for this test?

leochr (commented):
@idlewis Looks good. Just one minor question. Thanks.

utils/reconciler_test.go
idlewis: The test is testing the CreateOrUpdate function, so I don't think the the return value from CustomizeServiceAccount matters (the test could call any function at that point)
That said, I'm not sure the test is a particularly good one as it doesn't seem to verify the behaviour of CreateOrUpdate very much. But if that needs changing, it should probably have a separate issue.

leochr (approved):

Issue #340, opened 2022-03-01
Support pod topology spread constraints


leochr:
The operator supports affinity (node affinity, pod affinity, pod anti-affinity). To provide more control, consider supporting [Pod Topology Spread Constraints](https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/) (topologySpreadConstraints)

greglanthier:
I'm working on an operator that currently relies on **`topologySpreadConstraints`** to aid in scheduling workloads in a more fault tolerant manner.  I would like to transition to using RuntimeComponent CRs to manage the workloads my operator deploys rather than have my operator manage the workloads directly.

How do y'all envision **`topologySpreadConstraints`** would be added to the runtime-component-operator CRD?

Would this follow the same pattern of [`PodSpec`](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#PodSpec) - in the sense that a `topologySpreadConstraints` stanza would be a sibling of the `affinity` stanza?

Has this work been planned?  Is there any way I can help - either with PRs or with testing?

halim-lee:
Hi @greglanthier, the work has not been planned as of yet since it is an enhancement idea for the future. It will likely be added as `spec.topologySpreadConstraints` in the CR as `spec.affinity` is. I will update this issue as soon as I start implementing it!

greglanthier:
Thanks for the update @halim-lee.

Since our workloads currently rely on **`topologySpreadConstraints`** and we can't transition to runtime-component-operator until it supports I'd be happy to contribute - be that with code or by providing some testing.

I'll keep watching this space for updates.  Thanks again.

leochr:
@greglanthier Support for topology spread constraints is available in Liberty Operator and Runtime Component Operator 1.3.0+

Closing as delivered.

PR #410, opened 2022-07-12
Add Topology Spread Constraints support

halim-lee:

- Topology Spread Constraints support

- [ ] User guide
- [ ] `CHANGELOG.md`

Fixes #


halim-lee:
Related: https://github.com/application-stacks/runtime-component-operator/issues/340

utils/utils.go
```diff
@@ -738,6 +740,21 @@ func CustomizePodSpec(pts *corev1.PodTemplateSpec, ba common.BaseComponent) {
 	pts.Spec.AutomountServiceAccountToken = &mount
 }
 
+// Initialize an empty TopologySpreadConstraints list and optionally prefers scheduling across zones for pods with zoneDomainMatchLabels
+func CustomizeTopologySpreadConstraints(pts *corev1.PodTemplateSpec, zoneDomainMatchLabels map[string]string) {
+	pts.Spec.TopologySpreadConstraints = make([]corev1.TopologySpreadConstraint, 0)
+	if len(zoneDomainMatchLabels) > 0 {
+		pts.Spec.TopologySpreadConstraints = append(pts.Spec.TopologySpreadConstraints, corev1.TopologySpreadConstraint{
```
leochr, line 747: Should we add a similar constraint using topology key `kubernetes.io/hostname`? It'll find the most ideal node I believe as per this statement in the k8s doc:

"When a Pod defines more than one topologySpreadConstraint, those constraints are combined using a logical AND operation: the kube-scheduler looks for a node for the incoming Pod that satisfies all the configured constraints."

utils/utils.go
```diff
@@ -1215,6 +1232,25 @@ func GetWatchNamespaces() ([]string, error) {
 	return watchNamespaces, nil
 }
 
+// MergeTopologySpreadConstraints returns the union of all TopologySpreadConstraint lists. The order of lists passed into the
+// func, defines the importance.
+func MergeTopologySpreadConstraints(constraints ...[]corev1.TopologySpreadConstraint) []corev1.TopologySpreadConstraint {
```
leochr, line 1237: Customers could set [default constraints for the entire cluster](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#cluster-level-default-constraints). But those defaults would be overridden by our default constraints. It'll be good to consider providing an option/way, so they can disable our defaults without having to override each constraint (using the relevant topology key)

leochr (commented):
Thanks for the PR. Added some comments.

utils/utils.go
kabicin: Added flag `.spec.topologySpreadConstraints.disableOperatorDefaults`. It defaults to false, so operator default TopologySpreadConstraints will be applied if the flag isn't specified. However, this also means upgrading the operator will cause brief downtime for the pod to apply our default constraints.

utils/utils.go
kabicin: Yes, I think it's okay to add here. It's been documented to not always work the way we expect (because the scheduler adds one pod at a time). But in such cases, we are covered by the `ScheduleAnyway` flag and our pod anti-affinity rules.

api/v1/runtimecomponent_types.go
```diff
@@ -146,6 +146,21 @@ type RuntimeComponentSpec struct {
 	// Security context for the application container.
 	// +operator-sdk:csv:customresourcedefinitions:order=25,type=spec,displayName="Security Context"
 	SecurityContext *corev1.SecurityContext `json:"securityContext,omitempty"`
+
+	// Topology spread constraints for the application pod.
```
leochr, line 150: This line can be removed, so the comment above RuntimeComponentTopologySpreadConstraints type's struct can be used

leochr (commented):
@kabicin Thanks for addressing the review comments. Just one minor comment about the description.

api/v1/runtimecomponent_types.go
kabicin: Sounds good, I removed line 150.

leochr (approved):

PR #432, opened 2022-10-27
Refactor Issuer generation

arturdzm:


utils/reconciler.go
```diff
@@ -383,61 +447,13 @@ func (r *ReconcilerBase) GenerateSvcCertSecret(ba common.BaseComponent, prefix s
...
-				Name: prefix + "-self-signed",
-			}
-
-			duration, err := time.ParseDuration(common.Config[common.OpConfigCMCADuration])
-			if err != nil {
-				return err
+		err = r.GenerateCMIssuer(bao.GetNamespace(), prefix, CACommonName, operatorName)
+		if err !=nil {
+			if errors.Is(err, APIVersionNotFoundError) {
+				return false, nil
```
leochr, line 453: Should this be `return false, err` ? similar to how _Certificate_ is handled ([here](https://github.com/application-stacks/runtime-component-operator/blob/2227b1491c3d32106d467e78e19dee636d8387d3/utils/reconciler.go#L446))

leochr (commented):
@arturdzm Thanks for the PR. Look good. Just one question about error handling

utils/reconciler.go
InhexSTER: In the original code when certmanager.io API not available it didn't return an error so that reconciler goes on, either using  other method (OCP) or using byo certificate.

arturdzm: In the original code when certmanager.io API not available it didn't return an error so that reconciler goes on, either using other method (OCP) or using byo certificate.

leochr (approved):
@arturdzm Thanks for the clarification

PR #435, opened 2022-11-07
Add a way to byo CA/Issuer

arturdzm:


utils/reconciler.go
```diff
@@ -448,7 +469,7 @@ func (r *ReconcilerBase) GenerateSvcCertSecret(ba common.BaseComponent, prefix s
 		bao := ba.(metav1.Object)
 
 		err = r.GenerateCMIssuer(bao.GetNamespace(), prefix, CACommonName, operatorName)
```
leochr, line 471: Should this be skipped if a custom issuer exists?

leochr (commented):
@arturdzm Looks good. Added a question about the custom issuer. Thanks.

utils/reconciler.go
arturdzm: I decided to leave this on just so that it is always there,  and since default issuer / CA are not owned by application they will not be deleted with apps. 
We might want to revisit in the future, just didn't want to add complexity of cleanup at this point

leochr: Okay, let's revisit. Agree that we don't need to clean up the default issuer and its resources, but the method call GenerateCMIssuer can be skipped if a custom issuer is used.

PR #436, opened 2022-11-16
Issue326

smcclem:

Sub-operators using ManagerError will have the resource status overwritten in the event the application status is OK.

- [ ] User guide
- [ ] `CHANGELOG.md`

N/A
Fixes #


leochr (approved):

leochr:
@smcclem just noticed that the if-check on CheckApplicationStatus method call is added in ManageSuccess. But it should be checked in ManageError.

smcclem:
> @smcclem just noticed that the if-check on CheckApplicationStatus method call is added in ManageSuccess. But it should be checked in ManageError.

Yes, I was testing it with Kirby, and realized it before I saw your note here.  Moved the code and retested it with the latest changes.  Works now.

PR #460, opened 2023-02-21
Add dev script

smcclem:

Add dev.sh script, which adds some quick developer tools. It can build and push the operator to an OCP cluster, and install it with OLM (using the private registry). It can also setup and run the scorecard tests. 

- [ ] User guide
- [ ] `CHANGELOG.md`

Fixes #


Makefile
```diff
@@ -224,19 +253,19 @@ docker-login: ## Log in to a Docker registry.
...
-	docker push ${IMG}
+	$(CONTAINER_COMMAND) push  $(PODMAN_SKIP_TLS_VERIFY) ${IMG}
 
 .PHONY: bundle-build
 bundle-build: ## Build the bundle image.
-	docker build -f bundle.Dockerfile -t $(BUNDLE_IMG) .
+	$(CONTAINER_COMMAND) build -f bundle.Dockerfile -t $(BUNDLE_IMG) .
 
 .PHONY: bundle-push
-bundle-push: docker-login ## Push the bundle image.
```
idlewis, line 238: As the following command still might run docker, is it safe to remove the 'docker-login' dependency here?

scripts/dev.sh
```diff
@@ -0,0 +1,539 @@
+#!/bin/bash
```
idlewis, line 1: This doesn't seem to be checked in with execute permission. Might be helpful to do so.

Makefile
smcclem: Both WLO and OLO do not have this dependency here. I am not sure why its here and if there is an actual need for it.

idlewis (approved):
LGTM

Issue #299, opened 2021-12-16
Reset host in Route resource


leochr:
Configure host for Route:

```yaml
spec:
  expose: true
  route:
    host: mycompany-host
```

Created Route resource will get the host values

Then remove the route field
```yaml
spec:
  expose: true
```

The Route resource still has the host value.

- RCO
   - [x] Fixed
   - [x] Added tests
- OLO
   - [x] Fixed
   - [x] Added tests
- WLO
   - [x] Fixed
   - [x] Added tests

leochr:
Changing this issue for route host onlly. Something to address in the future. 
Route path not being removed issue is tracked/fixed via https://github.com/OpenLiberty/open-liberty-operator/issues/363

idlewis:
This isn't quite done as some code should be common.
#618 does this for RCO, but OLO and WLO are also needed

idlewis:
https://github.com/WASdev/websphere-liberty-operator/pull/631
https://github.com/OpenLiberty/open-liberty-operator/pull/567
Make the move to common code for OLO and WLO

leochr:
Closing as completed

PR #466, opened 2023-03-10
Always set the route path to empty if not specified

idlewis:
As otherwise a previous path may continue to exist in the route, even if it has been removed from the spec

This is to deal with part of the problem raised in issue #299
When this is merged, the OpenLiberty operator will need to pick up the latest utils, to fix the similar issue
OpenLiberty/open-liberty-operator#363


leochr (approved):
Looks good. Updated bundle/tests/scorecard/kuttl/routes/01-add-path.yaml to remove the old image (not compatible with P and Z). The CR update to add route path should still take effect.

PR #475, opened 2023-03-29
Handle k8s resource conflict due to same CR names

smcclem:

Checks to see if there is an existing Deployment, Statefulset or Knative service with the same name, and verifies
if it is managed by another operator.  If its not,  the status is updated to reflect the issue and it is re-queued. 

in addition to a summary of the change and link to the pull request.
-->
- [ ] User guide
- [ ] `CHANGELOG.md`

Fixes #  https://github.com/WASdev/websphere-liberty-operator/issues/179

### DIFF
--- a/controllers/runtimecomponent_controller.go
+++ b/controllers/runtimecomponent_controller.go
@@ -234,6 +234,13 @@ func (r *RuntimeComponentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		reqLogger.V(1).Info(fmt.Sprintf("%s is not supported on the cluster", servingv1.SchemeGroupVersion.String()))
 	}
 
+	// Check if there is an existing Deployment, Statefulset or Knative service by this name
+	// not managed by this operator
+	err = appstacksutils.CheckForExistingDeployments("RuntimeComponent", instance.Name, instance.Namespace, r.GetClient(), req, isKnativeSupported)
+	if err != nil {
+		return r.ManageError(err, common.StatusConditionTypeReconciled, instance)
+	}
+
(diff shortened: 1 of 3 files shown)


controllers/runtimecomponent_controller.go
```diff
@@ -234,6 +234,13 @@ func (r *RuntimeComponentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		reqLogger.V(1).Info(fmt.Sprintf("%s is not supported on the cluster", servingv1.SchemeGroupVersion.String()))
 	}
 
+	// Check if there is an existing Deployment, Statefulset or Knative service by this name
```
leochr, line 237: To avoid processing the service account above, perform this check earlier (suggest [here](https://github.com/application-stacks/runtime-component-operator/blob/671d91fff45bc97996a66891c659882bd7ad58a0/controllers/runtimecomponent_controller.go#L137))

controllers/runtimecomponent_controller.go
```diff
@@ -234,6 +234,13 @@ func (r *RuntimeComponentReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+	// not managed by this operator
+	err = appstacksutils.CheckForExistingDeployments("RuntimeComponent", instance.Name, instance.Namespace, r.GetClient(), req, isKnativeSupported)
```
leochr, line 239: Suggest name: `CheckForNameConflicts`

leochr (commented):
@smcclem Thanks for the PR. Looks good for the most part. Added couple of comments.

smcclem:
@leochr

leochr (approved):
@smcclem Thanks for the updates. Looks good

Issue #457, opened 2023-02-13
Update build and test scripts of RCO and OLO


leochr:
Blocked by: WLO issue https://github.com/WASdev/websphere-liberty-operator/issues/396

- [ ] Update build and test scripts to sync with WLO
- [ ] Clean-up/archive unused scripts, makefile instructions and functions

- [ ] Delivered to OLO

idlewis:
@leochr There are still some significant differences in the .one-pipeline.yaml script between RCO and the others. Apart from that I believe this is all done.

PR #476, opened 2023-03-30
Sync scripts and makefile with equivalent WLO files

idlewis:
For issue #457


leochr:
Included with https://github.com/application-stacks/runtime-component-operator/pull/485. Thanks

PR #477, opened 2023-03-30
changes to kuttl tests to make them multi-arch compatible

yongja79:

- the kuttl tests need updating to make them multi-arch compatible 
- Some tests will fail in the current automation as they have been updated to work with onepipeline.

- [ ] User guide
- [ ] `CHANGELOG.md`


leochr (approved):
@yongja79 Looks good. Thank you

PR #479, opened 2023-03-30
Check Issuer observed generation

arturdzm:
Make sure that certificate Issuer generation has been observed by cert-manager before using it


leochr (approved):
@arturdzm Looks good. Thank you

PR #484, opened 2023-04-04
Allow custom annotations on service Certificate

arturdzm:


leochr (approved):
@arturdzm Looks good. Updated the display name and description slightly. Thank you.

PR #496, opened 2023-04-12
Add version info into log

arturdzm:


leochr (commented):
We'll release RCO and OLO with version 1.2.0 to keep it sync with WLO. So version should be updated. Looks fine otherwise. Thanks

PR #501, opened 2023-04-14
Update kustomize files

yongja79:

- Move the kustomize files from deploy/kustomize/daily to internal/deploy/kustomize/daily and update any references
- Update permissions in kustomize role file
- Update the kusomize files to be inline with WLO operator
- Update Makefile to update kustomize files  when make bundle is called
- Update kustomize readme to reflect the change in command as kubectl create has to be used instead of apply due to size constraints

- [ ] User guide
- [ ] `CHANGELOG.md`


halim-lee (approved):
Looks good to me!

Issue #445, opened 2023-01-17
Comply with `restricted:v1.24`


halim-lee:
While running applications on OCP 4.11, operator complains that app container violates PodSecurity.
```
W0117 15:08:56.801191 1 warnings.go:67] would violate PodSecurity "restricted:v1.24": seccompProfile (pod or container "app" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
```

Set `securityContext.seccompProfile.type`

halim-lee:
https://kubernetes.io/docs/tutorials/security/seccomp/

halim-lee:
- RuntimeDefault: represents the default container runtime seccomp profile
- Localhost: indicates a profile defined in a file on the node should be used. The file's location relative to <kubelet-root-dir>/seccomp.

The default should be RuntimeDefault. See [here](https://kubernetes.io/docs/tutorials/security/seccomp/#create-pod-that-uses-the-container-runtime-default-seccomp-profile)
```Most container runtimes provide a sane set of default syscalls that are allowed or not. You can adopt these defaults for your workload by setting the seccomp type in the security context of a pod or container to RuntimeDefault.```

PR #508, opened 2023-04-19
Set RuntimeDefault as default for SeccompProfile

halim-lee:

- To comply with restricted:v1.24

- [ ] User guide
- [ ] `CHANGELOG.md`

Related: https://github.com/application-stacks/runtime-component-operator/issues/445

### DIFF
--- a/bundle/manifests/runtime-component.clusterserviceversion.yaml
+++ b/bundle/manifests/runtime-component.clusterserviceversion.yaml
@@ -71,7 +71,7 @@ metadata:
     categories: Application Runtime
     certified: "true"
     containerImage: icr.io/appcafe/runtime-component-operator:daily
-    createdAt: "2024-08-29T22:22:00Z"
+    createdAt: "2024-09-11T20:04:51Z"
     description: Deploys any runtime component with dynamic and auto-tuning configuration
     features.operators.openshift.io/disconnected: "true"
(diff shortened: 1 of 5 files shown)


halim-lee:
Tests passed in: https://cloud.ibm.com/devops/pipelines/tekton/729b3aea-0a81-4b4d-bd42-7e5811980010/runs/c7768291-7a90-4f39-bb37-e17f5287d7c9/deploy-acceptance-tests/run-stage?env_id=ibm:yp:us-south

leochr (commented):
@halim-lee Thank you for the PR. Looks good. Just added one comment. 

It'll be good to include this with the 1.4.0 release if possible.

OpenShift specifies a default seccomp policy by default, but wondering whether Kubernetes does. If not, in our kind cluster we should see a warning (or error if it has changed from warning).

halim-lee:
> OpenShift specifies a default seccomp policy by default, but wondering whether Kubernetes does. If not, in our kind cluster we should see a warning (or error if it has changed from warning).

From what I see here: https://kubernetes.io/docs/tutorials/security/seccomp/#create-a-pod-that-uses-the-container-runtime-default-seccomp-profile, RuntimeDefault is the default value they use for seccomp profile. If seccomp profile is not enabled, Unconfined (seccomp disabled) mode is the default.

Issue #506, opened 2023-04-19
Add labels to operator config map


idlewis:
The operator stores config a ConfigMap. When this is created, the following labels should be added:

"app.kubernetes.io/instance"
"app.kubernetes.io/managed-by"
"app.kubernetes.io/name"

This was noticed as being missing due to issue WASdev/websphere-liberty-operator#448

PR #512, opened 2023-04-20
Always use utils method to create controller configmap

idlewis:
This ensure the correct labels are added

Second half of the fix for #506

### DIFF
--- a/controllers/runtimecomponent_controller.go
+++ b/controllers/runtimecomponent_controller.go
@@ -25,7 +25,6 @@ import (
 	"github.com/application-stacks/runtime-component-operator/common"
 	"github.com/pkg/errors"
 
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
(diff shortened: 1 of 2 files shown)


controllers/runtimecomponent_controller.go
```diff
@@ -104,22 +103,11 @@ func (r *RuntimeComponentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	configMap, err := r.GetOpConfigMap(OperatorName, ns)
 	if err != nil {
 		reqLogger.Info("Failed to find runtime-component-operator config map")
-		common.Config = common.DefaultOpConfig()
-		configMap = &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: OperatorName, Namespace: ns}}
-		configMap.Data = common.Config
+		appstacksutils.CreateConfigMap(OperatorName)
 	} else {
 		common.Config.LoadFromConfigMap(configMap)
```
leochr, line 108: I think we should get rid of this `else` check and move this outside to always load the ConfigMap (i.e. it was just created and we need to load it)

utils/utils.go
```diff
@@ -1523,12 +1523,14 @@ func CreateConfigMap(mapName string) {
...
+		watchNamespaces, err := GetWatchNamespaces()
+		if err != nil {
+			utilsLog.Error(err, "Error getting watch namespace")
+			return
+		}
+		// If the operator is running locally, use the first namespace in the `watchNamespaces`
+		operatorNs = watchNamespaces[0]
 	}
 	configMap := &corev1.ConfigMap{}
 	err := client.Get(context.TODO(), types.NamespacedName{Name: mapName, Namespace: operatorNs}, configMap)
```
leochr, line 1536: It'll be good to check that the error was `IsNotFound` ([example](https://github.com/application-stacks/runtime-component-operator/blob/4ac56bcc9077dd495d87931cfb5bbd26539f2d50/utils/reconciler.go#L135)) and only proceed to create ConfigMap if so. Otherwise, we could update the ConfigMap with default settings and override any user-specified configuration when we encounter any error

So changing `if err != nil {` --> `if err != nil && apierrors.IsNotFound(err) {`

leochr (commented):
@idlewis Thanks for the PR. Added a couple of comments

controllers/runtimecomponent_controller.go
idlewis: I don't think I can do this without modifying appstacksutils.CreateConfigMap(OperatorName)
That function doesn't return the configmap at the moment, so it is possible that we could 'load' an empty map.
It is a bit yuck, but appstacksutils.CreateConfigMap does already copy the default config into common.Config, so I believe the current code is correct.
As previously discussed, I was trying to not change the signature of CreateConfigMap, hence why the code is the way it is. Happy to make further changes if you would prefer.

leochr: Ahh okay. Missed the default config being set on common.Config. We should be good then. Thanks

leochr (approved):

PR #530, opened 2023-05-02
Add Status.Versions.Reconciled to runtime components

idlewis:
This is to deal with issue #489

Issue #489, opened 2023-04-11
Add reconciled version


leochr:
Add status.versions.reconciled for RCO. 

WLO and OLO already include it

- [x] add checks to RuntimeOperations test in RCO
- [x] add equivalent tests to OLO
- [x] add equivalent tests to WLO

leochr:
@idlewis It seems WLO needs the tests (although WLO is checked off above)

idlewis:
> @idlewis It seems WLO needs the tests (although WLO is checked off above)

I clearly can't read, I thought I had done this.
PR opened for WLO is WASdev/websphere-liberty-operator#480

PR #539, opened 2023-05-12
Add versions.reconciled test for operations

idlewis:
As per issue #489 
follow up to pr #530


leochr (approved):

PR #548, opened 2023-05-30
Use ubi instead of golang image for main Dockerbuild

idlewis:
To deal with dockerhub rate limits 

See WASdev/websphere-liberty-operator#270

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
 # Build the manager binary
-FROM registry.access.redhat.com/ubi8 as builder
+FROM registry.access.redhat.com/ubi8-minimal:latest as builder
 ARG GO_PLATFORM=amd64
 env PATH=$PATH:/usr/local/go/bin
-RUN yum install -y wget
-RUN wget --no-verbose --header "Accept: application/octet-stream" "https://golang.org/dl/go1.20.10.linux-${GO_PLATFORM}.tar.gz" -O - | tar -xz -C /usr/local/
+RUN microdnf install tar gzip
+RUN curl -L --output - "https://golang.org/dl/go1.20.10.linux-${GO_PLATFORM}.tar.gz" | tar -xz -C /usr/local/
 
 WORKDIR /workspace
 # Copy the Go Modules manifests
@@ -58,3 +58,4 @@ COPY --from=builder /workspace/manager .
 USER ${USER_ID}:${GROUP_ID}
 
 ENTRYPOINT ["/manager"]
+
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM registry.access.redhat.com/ubi8 as builder
 ARG GO_PLATFORM=amd64
 env PATH=$PATH:/usr/local/go/bin
 RUN yum install -y wget
-RUN wget --no-verbose --header "Accept: application/octet-stream" "https://golang.org/dl/go1.20.linux-${GO_PLATFORM}.tar.gz" -O - | tar -xz -C /usr/local/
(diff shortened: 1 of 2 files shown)


halim-lee (approved):
Looks good!

Dockerfile
```diff
@@ -1,5 +1,9 @@
 # Build the manager binary
-FROM golang:1.20 as builder
+FROM registry.access.redhat.com/ubi8 as builder
```
leochr, line 2: Can `registry.access.redhat.com/ubi8/ubi-minimal:latest` be used instead?

leochr (commented):
@idlewis Wondering whether the operator binary from the new approach be compared with the current approach to ensure there isn't a significant difference

Dockerfile
idlewis: @leochr I've pushed some updates to use ubi-minimal.
I've tested locally and have a pipeline build running now.

idlewis:
> @idlewis Wondering whether the operator binary from the new approach be compared with the current approach to ensure there isn't a significant difference

@leochr I have had a look around and I can't find a method to do a meaningful compare.
`go version -m <go binary>`
lists the go version and modules used to compile a binary, that is the best I have found.

One issue the above command did show is that I was using go 1.20.0 rather than 1.20.10. It seems the golang 1.20 docker image uses the 'latest' 1.20 version of go. However, the go download URL for 1.20 seems to specifically get 1.20.0. I haven't found an alternative download URL. This would mean we would have to update the docker file regularly to keep up with minor version updates.

Dockerfile
```diff
@@ -1,5 +1,9 @@
 # Build the manager binary
-FROM golang:1.20 as builder
+FROM registry.access.redhat.com/ubi8-minimal:latest as builder
+ARG GO_PLATFORM=amd64
+env PATH=$PATH:/usr/local/go/bin
```
leochr, line 4: Should this be `ENV` instead of `env`?

Dockerfile
```diff
@@ -1,5 +1,9 @@
...
+RUN microdnf install tar gzip
+RUN curl -L --output - "https://golang.org/dl/go1.20.10.linux-${GO_PLATFORM}.tar.gz" | tar -xz -C /usr/local/
```
leochr, line 6: The Go version should be parameterized. Link it to the OnePipeline's config `go-version`

https://github.ibm.com/websphere/operators/blob/3f38eb6677e355cdc7a5320a0c2cb9690709301f/scripts/pipeline/containerize-stage.sh#L109

leochr (commented):
@idlewis Thanks for the PR. Added a few comments.

leochr:
@idlewis The instructions at https://go.dev/doc/install state to use 
`rm -rf /usr/local/go && tar -C /usr/local -xzf go1.21.3.linux-amd64.tar.gz`

It doesn't seem like UBI image includes `/usr/local/go` but it would be good to remove the directory.
Also wondering if the `-f` option should be used on `tar` command

Dockerfile
idlewis: Yes. It seems to work with 'env' but that isnt' documented, so I've updated it.

idlewis:
> @idlewis The instructions at https://go.dev/doc/install state to use `rm -rf /usr/local/go && tar -C /usr/local -xzf go1.21.3.linux-amd64.tar.gz`
> 
> It doesn't seem like UBI image includes `/usr/local/go` but it would be good to remove the directory. Also wondering if the `-f` option should be used on `tar` command

I've added the `rm` command.
The `-f` isn't needed in the tar command as we are extracting from stdin, instead of a file

leochr (approved):

PR #549, opened 2023-06-02
Remove test step timeouts shorter than overall timeout

idlewis:
The default test step timeout in kuttl-test.yaml is now set to 150s. This means that tests which were setting a step timeout of 60s were actually less robust to timing issues.


yongja79 (approved):
LGTM

Issue #538, opened 2023-05-11
Document pre-req for manageTLS: runtime must load certificates


leochr:
manageTLS mounts the certificates onto the runtime pods, but the runtime is responsible for loading the mounted certificates. Liberty containers load the certificates automatically into the default keystore / truststore at container startup. tWAS runtime doesn't have such equivalence. Document that the inorder for manageTLS to work end-to-end, runtime must load the certs.

tam512:
The current workaround to install tWAS applications using RCO v1.2.0 is to use `spec.route.termination: edge` with `insecureEdgeTerminationPolicy: Redirect` or `insecureEdgeTerminationPolicy: Allow` or `spec.route.termination: edge.passthrough`

For example:
```
spec:
  route:
    termination: edge
    insecureEdgeTerminationPolicy: Redirect
  service:
    type: ClusterIP
    port: 9080
```

PR #562, opened 2023-07-20
Add limitiations doc for manageTLS

idlewis:
Document that runtime components are responsible for making use of the TLS certificates created by the operator when mangeTLS=true

deals with issu #538


doc/user-guide-v1.adoc
```diff
@@ -343,3 +343,10 @@ NOTE: The `RuntimeOperation` CR must be created in the same namespace as the Pod
 === Troubleshooting
 
 See the link:++troubleshooting.adoc++[troubleshooting guide] for information on how to investigate and resolve deployment problems.
+
+== Limitations
+=== Using certificates when manageTLS=true
+If `manageTLS=true` is specified in the CR YAML file, then the operator will manage the creation of a TLS secret for the runtime component.
```
leochr, line 349: manageTLS is true by default - even if it's not specified in CR YAML. It'll be good to mention/clarify so that users don't remove it and expect manageTLS to be false

leochr (commented):
@idlewis Thanks for the PR. Added a comment about the default value of manageTLS.

It'll be good to further capture the use case with WebSphere Traditional applications (comment from Tam [here](https://github.com/application-stacks/runtime-component-operator/issues/538#issuecomment-1544475043))

idlewis:
> @idlewis Thanks for the PR. Added a comment about the default value of manageTLS.
> 
> It'll be good to further capture the use case with WebSphere Traditional applications (comment from Tam [here](https://github.com/application-stacks/runtime-component-operator/issues/538#issuecomment-1544475043))

@leochr I'm not quite sure how to use the comment from Tam. There seem to be two suggestions on how to work around this issue with tWAS, but I'm guessing that the correct solution would depend on the tWAS configuration.

Were you thinking of documenting a specific use case, such as:
"If your application has its own certificates and will terminate TLS traffic, then set 'manageTLS=false', and 'termination=passthrough' ?

leochr:
@idlewis Specific use cases would be more helpful. The [troubleshooting doc](https://github.com/application-stacks/runtime-component-operator/blob/main/doc/troubleshooting.adoc) would be more appropriate than the Limitations section I guess to hold such information. Thanks

PR #563, opened 2023-07-20
Add service certificate annotations Kuttl tests

kabicin:
- Adds service certificate annotations Kuttl tests to check that annotations can be properly applied to CertManager Certificates with RCO's `.spec.service.certificate.annotations`

Related issue: https://github.com/WASdev/websphere-liberty-operator/issues/440


kabicin:
Closing since it is merged into the common repo.

PR #573, opened 2023-09-06
Allow insecureEdgeTerminationPolicy to be set independently

idlewis:
Even if 'termination' is not set in the config

this if for issue OpenLiberty/open-liberty-operator#463

Tests added to the consolidated repo


kabicin (approved):
LGTM

PR #575, opened 2023-09-12
Add .spec.serviceAccount.mountToken config option

idlewis:
Deals with RCO portion of OpenLiberty/open-liberty-operator#462

### DIFF
--- a/bundle/manifests/rc.app.stacks_runtimecomponents.yaml
+++ b/bundle/manifests/rc.app.stacks_runtimecomponents.yaml
@@ -3762,10 +3762,24 @@ spec:
                       a service
                     type: string
                 type: object
+              serviceAccount:
+                description: The service account to use for deploying the application.
+                  A service account is automatically created if this is not specifed.
+                properties:
+                  mountToken:
+                    description: Whether the Service Account token should be mounted
+                      into the application pods. Defaults to true.
+                    type: boolean
+                  name:
+                    description: Name of the service account to use for deploying
+                      the application. A service account is automatically created
+                      if it's not specified.
+                    type: string
+                type: object
               serviceAccountName:
-                description: Name of the service account to use for deploying the
-                  application. A service account is automatically created if it's
-                  not specified.
+                description: Deprecated. .spec.serviceAccount.name should be used
+                  instead. If both are specified, .spec.serviceAccount.name will override
+                  this.
                 type: string
               sidecarContainers:
                 description: List of sidecar containers. These are additional containers
--- a/bundle/manifests/runtime-component.clusterserviceversion.yaml
+++ b/bundle/manifests/runtime-component.clusterserviceversion.yaml
@@ -11,12 +11,15 @@ metadata:
             "name": "runtimecomponent-sample"
           },
           "spec": {
-            "applicationImage": "icr.io/appcafe/open-liberty/samples/getting-started@sha256:6bf52d536e94d6aa6eaa3ca9639e0fe9f5e71dbd891c50362939dd6aca519d02",
+            "applicationImage": "icr.io/appcafe/open-liberty/samples/getting-started@sha256:f7c7da21059eef8734cf0d43a417609aecf68bfe89d0be8e61012fade5877a01",
             "expose": true,
             "manageTLS": true,
             "replicas": 1,
             "service": {
               "port": 9443
+            },
+            "serviceAccount": {
+              "mountToken": true
             }
           }
         },
@@ -41,7 +44,7 @@ metadata:
             "name": "runtimecomponent-sample"
           },
           "spec": {
-            "applicationImage": "icr.io/appcafe/open-liberty/samples/getting-started@sha256:6bf52d536e94d6aa6eaa3ca9639e0fe9f5e71dbd891c50362939dd6aca519d02",
+            "applicationImage": "icr.io/appcafe/open-liberty/samples/getting-started@sha256:f7c7da21059eef8734cf0d43a417609aecf68bfe89d0be8e61012fade5877a01",
             "expose": true,
             "replicas": 1,
             "service": {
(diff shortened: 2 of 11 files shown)


leochr (commented):
@idlewis Thanks for the PR. I'll suggest creating a new config section for Service Account to group relevant fields. With the addition of the mount option, we'll have 2, but any config related to the service account can be added here in the future:

```
spec:
   serviceAccount:
      name:
      mountToken:
```

The existing field `.spec.serviceAccountName` should be marked as deprecated. But we should look it up if `.spec.serviceAccount.name` is not set.

api/v1/runtimecomponent_types.go
```diff
@@ -144,6 +148,16 @@ type RuntimeComponentSpec struct {
 	SecurityContext *corev1.SecurityContext `json:"securityContext,omitempty"`
 }
```
leochr, line 150: Description is needed here. It'll be shown in OpenShift UI for `serviceAccount` field.
Suggestion: `Defines the service account`

api/v1/runtimecomponent_types.go
```diff
@@ -84,6 +84,10 @@ type RuntimeComponentSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:order=12,type=spec,displayName="Probes"
 	Probes *RuntimeComponentProbes `json:"probes,omitempty"`
 
+	// TODO fix the ordering
```
leochr, line 87: ServiceAccountName field's description (in API) should be marked as deprecated. `.spec.serviceAccount.name` overrides this if both are specified.

utils/utils.go
```diff
@@ -809,6 +809,17 @@ func CustomizeServiceAccount(sa *corev1.ServiceAccount, ba common.BaseComponent,
...
+	mount := true
+	basa := ba.GetServiceAccount()
+	if basa != nil {
+		if basa.GetMountSAToken() != nil && !*basa.GetMountSAToken() {
+			// not nil and set to false
+			mount = false
+		}
+	}
+	sa.AutomountServiceAccountToken = &mount
+
```
leochr, line 822: We should set `automountServiceAccountToken: false` on Pod as well. This way, user can disable mount even for the non-default SA.

leochr (commented):
@idlewis Thanks for the updates. Added some comments. 

We should add some tests as well. Fine to do that separately after this PR

utils/utils.go
idlewis: @leochr 
Can I clarify this?
Is the suggestion that we will have a single config item (.spec.ServiceAccount.MountToken), and if that is set to false, then we will set both 
 * AutomountServiceAccountToken=false on the service account we create
 * automountServiceAccountToken=false on the deployment pod spec
 
Alternatively we could just set this _only_ on the pod spec. That should cover both the case of using our service account or a user specified one, and would make the code simpler.

leochr: @idlewis It's fine to set it only on the pod spec (as it takes precedence over SA)

idlewis:
@leochr This is now ready for review

api/v1/runtimecomponent_types.go
```diff
@@ -54,10 +54,15 @@ type RuntimeComponentSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:order=5,type=spec,displayName="Pull Secret",xDescriptors="urn:alm:descriptor:io.kubernetes:Secret"
 	PullSecret *string `json:"pullSecret,omitempty"`
 
+	// Deprecated. .spec.serviceAccount.name should be used instead. If both are specified, .spec.serviceAccount.name will override this.
 	// Name of the service account to use for deploying the application. A service account is automatically created if it's not specified.
```
leochr, line 58: The second line of the description should be removed

api/v1/runtimecomponent_types.go
```diff
@@ -54,10 +54,15 @@ type RuntimeComponentSpec struct {
...
 	// +operator-sdk:csv:customresourcedefinitions:order=6,type=spec,displayName="Service Account Name",xDescriptors="urn:alm:descriptor:com.tectonic.ui:text"
```
leochr, line 59: I think it'll be good to hide this deprecated field from the OpenShift UI

api/v1/runtimecomponent_types.go
```diff
@@ -54,10 +54,15 @@ type RuntimeComponentSpec struct {
...
 	ServiceAccountName *string `json:"serviceAccountName,omitempty"`
 
+	// The service account to use for deploying the application. A service account is automatically created if this is not specifed
```
leochr, line 62: Need `.` at the end to be consistent

api/v1/runtimecomponent_types.go
```diff
@@ -144,6 +149,17 @@ type RuntimeComponentSpec struct {
 	SecurityContext *corev1.SecurityContext `json:"securityContext,omitempty"`
 }
 
+// Defines the service account
+type RuntimeComponentServiceAccount struct {
+	// Whether the Service Account token should be mounted into the application pods. Default true
```
leochr, line 154: To be consistent with some of the other default booleans:
`Default true` --> `Defaults to true.`

api/v1/runtimecomponent_types.go
```diff
@@ -144,6 +149,17 @@ type RuntimeComponentSpec struct {
...
+	// +operator-sdk:csv:customresourcedefinitions:order=1,type=spec,displayName="Auto mount Service Account Token",xDescriptors="urn:alm:descriptor:com.tectonic.ui:booleanSwitch"
```
leochr, line 155: Auto mount Service Account Token --> Mount Service Account Token

leochr (commented):
@idlewis Thanks for the updates. Looks good. Added some minor comments about descriptions and UI rendering.

idlewis:
@leochr I've addressed all the review comment.

api/v1/runtimecomponent_types.go
leochr: The period `.` at the end is needed

leochr (approved):

leochr:
@idlewis Thanks for the updates. Just a minor update to add a period is needed (see [comment](https://github.com/application-stacks/runtime-component-operator/pull/575#discussion_r1355808978)). Looks good otherwise. Added approval.

PR #578, opened 2023-09-29
Add .spec.podSecurityContext flag

kabicin:
Allows using `.spec.podSecurityContext` to configure the security context of the pod


kabicin:
Will use https://github.com/application-stacks/runtime-component-operator/pull/579/files to address the issue instead.

PR #583, opened 2023-10-20
Add ARM step to runtime-component-operator-build.yml

yongja79:

- In order to get RCO to build on ARM we need to have the orchestrate configured for Jekins to pick up


halim-lee (approved):
Looks good!

PR #587, opened 2023-10-30
Add service annotations for topology aware routing

kabicin:


leochr (approved):
@kabicin Looks good. Thanks

PR #592, opened 2023-11-10
Remove deprecated kustomize keys from install files

idlewis:
And split some patches up to one per file, as this is required by some versions of kustomize

RCO part of this issue:
https://github.com/OpenLiberty/open-liberty-operator/issues/460


internal/deploy/kustomize/daily/examples/watch-all-namespaces/kustomization.yaml
```diff
@@ -3,8 +3,9 @@ kind: Kustomization
...
 
-bases:
-- ../../overlays/watch-all-namespaces
 
-patchesStrategicMerge:
-- rco-ns.yaml
+resources:
+- ../../overlays/watch-all-namespaces
+patches:
+- path: role1.yaml
```
leochr, line 10: @idlewis Thanks for the PR. It'll be good to have file names that indicate what it contains. Perhaps the name of the resource could be used. i.e. rco-leader-election-cluster-rolebinding.yaml

idlewis:
@leochr @kabicin I've updated the files as per Leo's comments

leochr:
@idlewis Thanks for updating the file names. Please update OLO and WLO with the same when you get a chance.

@kabicin Thanks for the review and testing.

PR #607, opened 2024-01-30
Remove unneeded packages now that docker isn't needed

idlewis:
Helps with removing ubuntu base image


idlewis:
Backing off on this PR, as it may be better to move the dockerfile to the common build repo for all three operators.

Issue #606, opened 2024-01-26
Created default NetworkPolicies do not work when service port is different then target port


TomVgw:
### What did you do?

We have upgrade our OpenLibertyOperator.  Due to the upgrade, the operator started creating Network Policies.  
For the creation of the Network policies, it uses utils from the runtime-component-operator.
Creation is done [here](https://github.com/OpenLiberty/open-liberty-operator/blob/699fcaa253d30ece5b2c66d0f2ffc1d89806bd7b/controllers/openlibertyapplication_controller.go#L383), function definition [here](https://github.com/OpenLiberty/open-liberty-operator/blob/699fcaa253d30ece5b2c66d0f2ffc1d89806bd7b/controllers/openlibertydump_controller.go#L9).

The specs of the service that was created by the operator looks like this:
`spec:
  clusterIP: 10.141.199.72
  ipFamilies:
    - IPv4
  ports:
    - name: http
      protocol: TCP
      port: 80
      targetPort: 9080
    - name: https
      protocol: TCP
      port: 443
      targetPort: 9443
  internalTrafficPolicy: Cluster
  clusterIPs:
    - 10.141.199.72
  type: ClusterIP
  ipFamilyPolicy: SingleStack
  sessionAffinity: None
  selector:
    app.kubernetes.io/instance: organisatiestructuur-service-dev`

### What did you expect to see?

We expected the Networkpolicies to allow ingress traffic., so no impact to a route defined within the OLA.
Within the network policies being created, we expected to find the ports of the pods, not the ports of the service.

We expected the following  code:
`spec:
  podSelector:
    matchLabels:
      apps.openliberty.io/name: organisatiestructuur-service-dev
  ingress:
    - ports:
        - protocol: TCP
          port: 9080
        - protocol: TCP
          port: 94433
      from:
        - namespaceSelector:
            matchLabels:
              policy-group.network.openshift.io/ingress: ''
        - namespaceSelector:
            matchLabels:
              network.openshift.io/policy-group: ingress
        - podSelector:
            matchLabels:
              app.kubernetes.io/part-of: organisatiestructuur-service-dev
          namespaceSelector:
            matchLabels:
              kubernetes.io/metadata.name: system-development-dev
        - namespaceSelector:
            matchLabels:
              network.openshift.io/policy-group: monitoring
  policyTypes:
    - Ingress`

### What did you see instead?

The created network policy mentions the port of the service.  Therefor, the defined routes failed.  Created network policy rules:
[...]

PR #612, opened 2024-02-29
NetworkPolicy to use targetPort and fallback on service port

kabicin:
Resolves issue in #606.

This change addresses the default NetworkPolicy for the RuntimeComponent, so that the ingress port matches the `.spec.service.targetPort` (pod's port) and uses the service's port `.spec.service.port` as a fallback, when the `targetPort` for the Pod is not specified.

In the previous code, the ingress port for the NetworkPolicy only matched the service `port`. When the pod's port `targetPort` is not the same as the service `port`, traffic to the Pod did not get covered by the NetworkPolicy.


utils/utils.go
```diff
@@ -448,9 +448,13 @@ func createNetworkPolicyPeer(appName string, namespace string, networkPolicy com
...
 	var ports []int32
-	ports = append(ports, ba.GetService().GetPort())
+	primaryTargetPort := ba.GetService().GetPort()
+	if ba.GetService().GetTargetPort() != nil {
+		primaryTargetPort = *ba.GetService().GetTargetPort()
+	}
+	ports = append(ports, primaryTargetPort)
 	for _, port := range ba.GetService().GetPorts() {
-		ports = append(ports, port.Port)
+		ports = append(ports, int32(port.TargetPort.IntValue()))
```
arturdzm, line 457: NetworkPolicy supports named ports, so we can youse `port.TagetPort` as is without getting int value.
Our CRD I think can have something like this:

```
ports:
- targetPort: http
````
Above code I think would fail.  Also can you make sure `port.TargetPort` is always set? in spec its marked as optional. Might be good to wrap into nil check even if the api always sets targetPort at the moment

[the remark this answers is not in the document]
Comment (kabicin) [reply utils/utils.go]: In case the api doesn't set targetPort from port, I added an else block to fallback onto `port.Port` if TargetPort is not specified.

Comment (kabicin) [reply utils/utils.go]: Thanks, I reorganized the code to use an `intstr.IntOrString` type allowing me to just use `port.TargetPort`. Also, I added an empty string check on the target port.

File: utils/utils.go [hunk 2]
```diff
@@ -462,15 +457,29 @@ func customizeNetworkPolicyPorts(ingress *networkingv1.NetworkPolicyIngressRule,
...
+	if ba.GetService().GetTargetPort() != nil {
+		primaryPortIntVal = *ba.GetService().GetTargetPort()
+	}
+	primaryPort := intstr.IntOrString{Type: intstr.Int, IntVal: primaryPortIntVal}
+	ingress.Ports[0].Port = &primaryPort
+
+	// Initialize additional ports
+	for i := 0; i < len(ba.GetService().GetPorts()); i++ {
+		portPtr := &ba.GetService().GetPorts()[i]
+		if portPtr.TargetPort.String() != "" {
```
leochr, line 477: `String()` function could return `"<nil>"`, not sure whether that's something we should check for

leochr (commented):
@kabicin Thanks for the updates. Added just one comment/clarification. 

@arturdzm Could you please take a look at the updates? Thanks

utils/utils.go
kabicin: I think the TargetPort property is not a pointer https://pkg.go.dev/k8s.io/api/core/v1#ServicePort so it would have a default value initialized for elements of the struct.

arturdzm: I think that's correct, ServicePort.TargetPort  is not a point so it can not be nil (defaults to 0, Int type) it might be initialized to equal port.  It might be that `if portPtr.TargetPort.String() != ""` might be always `true` though.

leochr: Okay. Sounds good. I'll go ahead and merge. Thank you

leochr (approved):

PR #613, opened 2024-03-01
Removed unused args for e2e-ocp.sh

idlewis:
Remove args that are no longer required due to script changes


idlewis:
closign in favour of #619

PR #619, opened 2024-03-21
Removed unused args for e2e-ocp.sh

idlewis:
For clarity


yongja79 (commented):
```--channel "${DEFAULT_CHANNEL}"``` should probably remain as we do use that value in the e2e-ocp.sh script and so if we pass it in its easier to trace

yongja79 (approved):
LGTM

PR #641, opened 2024-07-09
Don't try to create certificates with CN>64bytes

idlewis:
These are rejected by certmanager as they are not allowed in the spec


idlewis:
@leochr This is ready for review. It passes local testing, but no e2e tests delivered yet.

leochr (approved):

PR #664, opened 2024-11-11
Add debug logging levels

idlewis:
To help reduce the amount of logging done in the reconcile loop

Partially addresses https://github.com/OpenLiberty/open-liberty-operator/issues/638

Adds configurable logging levels, and moves the 'Reconciled...' messages into a debug level to reduce the amount of logging produced in the default configuration.


utils/reconciler.go
```diff
@@ -106,6 +106,8 @@ func (r *ReconcilerBase) SetDiscoveryClient(discovery discovery.DiscoveryInterfa
 }
 
 var log = logf.Log.WithName("utils")
+// Level 1 corresponds to 'DEBUG'
+var logv1 = log.V(1)
```
leochr, line 110: Assuming this log level can be set at Operator runtime, we can support users providing the log level in the Operator config map as a next step.

utils/reconciler.go
```diff
@@ -122,7 +124,7 @@ func (r *ReconcilerBase) CreateOrUpdate(obj client.Object, owner metav1.Object,
 	var gvk schema.GroupVersionKind
 	gvk, err = apiutil.GVKForObject(obj, r.scheme)
 	if err == nil {
-		log.Info("Reconciled", "Kind", gvk.Kind, "Namespace", obj.GetNamespace(), "Name", obj.GetName(), "Status", result)
+		logv1.Info("Reconciled", "Kind", gvk.Kind, "Namespace", obj.GetNamespace(), "Name", obj.GetName(), "Status", result)
```
leochr, line 127: How many log levels are there?

Logs such as this was polluting the log file, so we should only log the overall reconcile entry and exit as Info and these ones for debug level.

leochr (commented):
@idlewis Thanks for the PR. It's looking promising. Added some questions/comments.

idlewis:
@leochr This is now fully working, could you take another look?

It is not currently possible to update the log level dynamically, because the operator only reads config in at startup.
I am sure that can be fixed, but it would be a non trivial update.

common/config.go
```diff
@@ -17,6 +18,28 @@ const (
 
 	// OpConfigCMCADuration default duration for cert-manager issued service certificate
 	OpConfigCMCertDuration = "certManagerCertDuration"
+
+	// OpConfigLogLevel the level of logs to be written
+	OpConfigLogLevel = "logLevel"
```
leochr, line 23: Suggest `operatorLogLevel` to avoid any confusion with app log level

common/config.go
```diff
@@ -17,6 +18,28 @@ const (
...
+
+	// The allowed values for OpConfigLogLevel
+	logLevelWarning  = "warning"
+	logLevelInfo     = "info"
+	logLevelDebug    = "debug"
```
leochr, line 28: On second thought about the levels, I think naming similar to [Liberty](https://openliberty.io/docs/latest/log-trace-configuration.html) would be better. Suggest `warning`, `info`, `fine`, `finer` and `finest`

leochr (commented):
@idlewis Thanks for the updates. Looks good. Added some comments. 
With the log level configuration in Operator's ConfigMap, we must support dynamic updates. Configuration via env variable in Subscription would be more suited if it can only be processed at startup.

idlewis:
Please ignore the comments about dynamic updates. I'd completely misunderstood.
With the current code, logging output is updated dynamically when the operator config map is updated.

leochr (approved):

PR #683, opened 2025-02-12
Optimize cert generation by delaying create/update requests

kabicin:

- Speeds up the time to load certificates by waiting for Secrets to exist in the cluster before requesting the remaining certificate chain.
- As demonstrated in past test runs, overwhelming cert manager with certificate create request when dependent resources are not yet accessible leads to slowing down of the cert generation process. This will help limit frequent re-queuing for cert manager operator.

- [ ] User guide
- [ ] `CHANGELOG.md`

Fixes #


leochr (approved):
@kabicin Looks good. Thank you

PR #702, opened 2025-05-14
Support Service session affinity in spec

kabicin:


leochr (approved):
@kabicin Thank you for the updates. Looks good.

PR #706, opened 2025-05-28
Allow skipping pull secret validation

idlewis:
part of https://github.com/OpenLiberty/open-liberty-operator/issues/726

### DIFF
--- a/api/v1/runtimecomponent_types.go
+++ b/api/v1/runtimecomponent_types.go
@@ -197,7 +197,7 @@ type RuntimeComponentServiceAccount struct {
 	// +operator-sdk:csv:customresourcedefinitions:order=2,type=spec,displayName="Service Account Name",xDescriptors="urn:alm:descriptor:com.tectonic.ui:text"
 	Name *string `json:"name,omitempty"`
 
-	// Don't verify that the service account has a valid pull secret. Defaults to false
+	// Skip verifying that the service account has a valid pull secret. Defaults to false.
 	// +operator-sdk:csv:customresourcedefinitions:order=3,type=spec,displayName="Skip service account pull secret validation",xDescriptors="urn:alm:descriptor:com.tectonic.ui:booleanSwitch"
 	SkipPullSecretValidation *bool `json:"skipPullSecretValidation,omitempty"`
 }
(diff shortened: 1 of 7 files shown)


idlewis:
@leochr Could you see if you happy with the approach here?
Still need to tidy up, add tests etc.

leochr (commented):
@idlewis Thank you for the PR. Few comments:
- As the configuration is specific to pull secret referenced in a service account, suggest providing the config within `.spec.serviceAccount.*` field.
- The toggle shouldn't control/skip the validation of the default service account that we create. It should only apply to custom service accounts.

idlewis:
@leochr I've update the changes, could you take another look?

leochr (commented):
@idlewis Thank you for the update. Adding the skip validation logic in `ServiceAccountPullSecretExists` function is good. Setting the `StatusReferenceSAResourceVersion` in status (at the end of the function) shouldn't be skipped when pull secret validation is skipped. The reference is needed to restart the pods when a change in the ServiceAccount is detected.

idlewis:
Tests are here:
https://github.ibm.com/websphere/operators/pull/244

idlewis:
@leochr I now have tests passing for this issue, so the PR is ready for review.

api/v1/runtimecomponent_types.go
```diff
@@ -195,6 +195,10 @@ type RuntimeComponentServiceAccount struct {
 	// Name of the service account to use for deploying the application. A service account is automatically created if it's not specified.
 	// +operator-sdk:csv:customresourcedefinitions:order=2,type=spec,displayName="Service Account Name",xDescriptors="urn:alm:descriptor:com.tectonic.ui:text"
 	Name *string `json:"name,omitempty"`
+
+	// Don't verify that the service account has a valid pull secret. Defaults to false
```
leochr, line 199: Suggestion: `Skip verifying that the service account has a valid pull secret. Defaults to false.`

leochr (commented):
@idlewis Thanks for the updates. Looks good. Just one minor suggestion. Please rebase. Let's aim to merge RCO changes on Thursday; WLO and OLO changes on Friday. Thank you.

PR #735, opened 2025-11-25
Set probes port and scheme based on .spec.manageTLS

kabicin:

- The port/scheme will default based upon the CR instance's `.spec.manageTLS` value. For example, if manageTLS is true or undefined, scheme will be HTTPS and otherwise, HTTP.

- [ ] User guide
- [ ] `CHANGELOG.md`

Fixes #


api/v1/runtimecomponent_types.go
```diff
@@ -765,6 +765,17 @@ func (cr *RuntimeComponent) GetManageTLS() *bool {
 	return cr.Spec.ManageTLS
 }
 
+func (cr *RuntimeComponent) GetManagedPort() int {
+	return int(cr.GetService().GetPort())
```
leochr: Validate that the default value Operator sets (when user hasn't configured Service port) is retrieved here.

kabicin: Added nil check on `GetService()` and fallback to `8080`

common/types.go
```diff
@@ -195,6 +196,90 @@ type BaseComponentStatefulSet interface {
 	GetAnnotations() map[string]string
 }
 
+// This struct is taken from the Probe specification in https://github.com/kubernetes/api/blob/v0.33.4/core/v1/types.go
+// +kubebuilder:object:generate=true
+type BaseComponentProbe struct {
```
leochr: This can be removed

common/common.go
```diff
@@ -59,3 +59,55 @@ func GetDefaultMicroProfileLivenessProbe(ba BaseComponent) *corev1.Probe {
...
+		probe.ProbeHandler.Exec = config.ProbeHandler.Exec
+	}
+	if config.ProbeHandler.GRPC != nil {
+		probe.ProbeHandler.GRPC = config.ProbeHandler.GRPC
+	}
+	if config.ProbeHandler.HTTPGet != nil {
+		if probe.ProbeHandler.HTTPGet == nil {
+			probe.ProbeHandler.HTTPGet = &corev1.HTTPGetAction{}
+		}
+		if config.ProbeHandler.HTTPGet.Port.Type != 0 {
```
leochr: Just wanted to confirm that a nil check on `config.ProbeHandler.HTTPGet.Port` is not needed here.

leochr (commented):
@kabicin Thank you for the updates. Added some comments and a clarification.

common/common.go
```diff
@@ -11,8 +11,8 @@ func GetDefaultMicroProfileStartupProbe(ba BaseComponent) *corev1.Probe {
 		ProbeHandler: corev1.ProbeHandler{
 			HTTPGet: &corev1.HTTPGetAction{
 				Path:   "/health/started",
-				Port:   intstr.FromInt(int(ba.GetService().GetPort())),
-				Scheme: "HTTPS",
+				Port:   intstr.FromInt(ba.GetManagedPort()),
+				Scheme: corev1.URISchemeHTTPS,
```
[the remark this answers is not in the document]
kabicin: Sure, some tests assume scheme is not affected by `.spec.manageTLS` so there may be some breaking changes. For example, some `probes` tests need adjustment after this change. Please see adjustments at   https://github.ibm.com/websphere/operators/pull/258/files#diff-d8a058f0da1d4a6c12d831cf53afe614d6e03adecc6b633a4a387d4294f239e2 for added context.

leochr (approved):
@kabicin Looks good. Thank you.

PR #748, opened 2026-02-06
Fix override for TLS_DIR and sync volumeMount path

### LINKED ISSUE #745: Unable to override `TLS_DIR` using `.spec.env`
### What did you do?

Setting a custom directory for `TLS_DIR` will cause a duplicate definition in the Pod that favours the default `/etc/x509/certs` instead of `/other/directory`. 
```
spec:
  env:
    - name: TLS_DIR
       value: /other/directory
```

Setting a custom directory for `TLS_DIR` should set it to `/other/directory`.

### What did you see instead?

Pod shows `TLS_DIR` as `/etc/x509/certs`

### Possible solution

This line should be ignored when `appContainer.Env` already contains `TLS_DIR`
https://github.com/application-stacks/runtime-component-operator/blob/e673c1d851acf6a57ffde6ec758ca2a7a95ff4b4/utils/utils.go#L689.
The change will also involve passing the env value here https://github.com/application-stacks/runtime-component-operator/blob/e673c1d851acf6a57ffde6ec758ca2a7a95ff4b4/utils/utils.go#L700

miuponn:

PR for #745 
- Ignores setting the default `TLS_DIR` if already defined in `.spec.env` and sets volume mount path to the overridden value.  

- [ ] User guide
- [ ] `CHANGELOG.md`


kabicin:
Hi @miuponn, can you also please rebase and squash the commits once the comments are addressed?

leochr (approved):

Issue #367, opened 2022-04-06
Adding volumes to StatefulSet after the initial creation


leochr:
Kubernetes StatefulSet itself doesn't allow volumes to be added to StatefulSets after their initial creation. 

Trying to add a volume, results in this error:
`Forbidden: updates to statefulset spec for fields other than 'replicas',
        'template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and
        'minReadySeconds' are forbidden`
        
 The same limitation is encountered via operator. 
- create StatefulSet by setting:
```
spec:
  statefulSet: {}
```

- try to add a volume to it
```
spec.
  statefulSet: 
    storage:
      size: 100Mi
      mountPath: /logs
```

Tasks:
- [ ] understand the reasons behind StatefulSet not supporting
- [ ] if it's not an anti-pattern to add volume afterwards, support adding it via operator. StatefulSet probably have to be deleted and recreated (meaning there will be some downtime)

halim-lee:
I would recommend adding support only for adding volumes for now. Since making changes to StatefulSet volumes also requires deletion of pods, I am not sure if it is safe to make that too available. If we decide to support modifying the existing volumes, we can implement it later.

PR #749, opened 2026-02-13
Add PriorityClass support for pod scheduling and preemption

miuponn:

PR for [#367](https://github.com/WASdev/websphere-liberty-operator/issues/367)
  - Adds optional `priorityClassName` field to `RuntimeComponentSpec` to allow users to assign k8s PriorityClass to operator-managed pods
  -  Also propagates `priorityClassName` to the Knative service spec when createKnativeService is enabled

PR for [OLO](https://github.com/OpenLiberty/open-liberty-operator/pull/852) [WLO](https://github.com/WASdev/websphere-liberty-operator/pull/853)
 
- [x] User guide
- [ ] `CHANGELOG.md`


api/v1/runtimecomponent_types.go
```diff
@@ -168,6 +168,10 @@ type RuntimeComponentSpec struct {
 	// The list of hostnames and IPs that will be injected into the application pod's hosts file
 	// +operator-sdk:csv:customresourcedefinitions:order=30,type=spec,displayName="Host Aliases"
 	HostAliases []corev1.HostAlias `json:"hostAliases,omitempty"`
+
+	// Name of the PriorityClass for the pod.
```
leochr: Suggestion: Name of the PriorityClass for the application pods.

utils/utils.go
```diff
@@ -765,6 +765,11 @@ func CustomizePodSpec(pts *corev1.PodTemplateSpec, ba common.BaseComponent) {
 	}
 
 	pts.Spec.Tolerations = ba.GetTolerations()
+
+	pts.Spec.PriorityClassName = ""
```
leochr: This could be added in an 'else' block to avoid always setting to empty (even when PriorityClass is defined).

leochr (commented):
@miuponn Thank you for the PR. Looks good. Just add couple of minor comments.

utils/utils.go
miuponn: added as else block, changed for knative as well

api/v1/runtimecomponent_types.go
miuponn: fixed

leochr (approved):
@miuponn Looks good. Thank you.

PR #766, opened 2026-04-09
Clear MP default httpGet probe handler

kabicin:

- Fixes issue with MicroProfile default httpGet handler not being cleared when user configures a probe handler in the CR spec. 
- Since only one probe handler should be set anyways, we can exit early after processing the first probe handler configured in the CR spec using `else if` checks.

- [ ] User guide
- [ ] `CHANGELOG.md`

Fixes #


leochr (approved):
@kabicin Looks good. Thank you.

